PR #13, opened 2016-11-07
Fix typo

chalasr:


chalasr:
Shall we keep coveralls flooding each PR? :)

ogizanagi:
Thank you @chalasr 

> Shall we let coveralls flooding each PR? :)

Of course not. It was enabled by default. I've disabled it now :)

PR #24, opened 2016-11-21
Add Symfony Validator Enum constraint

ogizanagi:
For flagged enums, it'll require another constraint and validator simply asking for the enum class and calling `accepts($value)` when validating a raw value (you can't create a flagged enum instance with an invalid value), but it's probably not worth it.

For other usages with flagged enums (as instances or a set of restricted values), this constraint should be enough.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,8 @@ To use the constraint, simply provide the enum `class`:
 # src/AppBundle/Resources/config/validation.yml
 AppBundle\Entity\User:
     properties:
-        gender: MyApp\Enum\Gender
+        gender:
+            - Elao\Enum\Bridge\Symfony\Validator\Constraint\Enum: MyApp\Enum\Gender
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -597,7 +597,9 @@ To use the constraint, simply provide the enum `class`:
 # src/AppBundle/Resources/config/validation.yml
 AppBundle\Entity\User:
     properties:
-        gender: MyApp\Enum\Gender
+        gender:
+            - Elao\Enum\Bridge\Symfony\Validator\Constraint\Enum:
+                class: MyApp\Enum\Gender
```
ogizanagi: class is the default value. For this particular sample: should be:

```diff
  AppBundle\Entity\User:
      properties:
         gender:
             - Elao\Enum\Bridge\Symfony\Validator\Constraint\Enum: MyApp\Enum\Gender
```

ogizanagi:
Thanks @benji07 for the hand up.

Issue #5, opened 2016-11-06
Consider using ReflectionClass::getConstants


ogizanagi:
I've always hesitated, but we can consider using [this function](http://php.net/manual/fr/reflectionclass.getconstants.php) in order to get the full list of defined constants in the given enumeration type.

Benefits:
- No need to implement `EnumInterface::values` in each enum type.
- Reflects the `\SplEnum` behavior and enumerations in other languages.
- ? :/

Drawbacks:
- Magic
- Cannot define other constants not meant to be used as an enum value. In theory, I'd say it does not matter much, as the only constants that should live inside the enum should be the ones defining enum values... But... there is the case of the `FlaggedEnum` which can also define common flags combinations, whereas only single bit flags are valid values. Can be workaround for this particular case, though.
- The previous point can be workaround by using naming conventions. But naming conventions are a drawback on their own IMHO.

chalasr:
So we could provide a `getPossibleValues()` which would makes use of reflection? (sorry if something similar is already implemented, I can't find it?)?

> Cannot define other constants not meant to be used as an enum value.

People can still override the method which makes use of the function, right?

ogizanagi:
> So we could provide a getPossibleValues() which would makes use of reflection? (sorry if something similar is already implemented, I can't find it?)?

Yes, the default implementation of `getPossibleValues()` (now renamed to `values` (simplified static API)) will use reflection and cache the result for subsequent calls.

> People can still override the method which makes use of the function, right?

Sure. But I'm not sure the idea of letting the user deals with it and have some enumerations behaving differently pleases me. Enumerations should work the same in the big lines IMHO.

But that's really something to consider.

chalasr:
> Enumerations should work the same in the big lines IMHO.

I agree. Considering

> - Cannot define other constants

and

> - Magic

I fear that using this function may cause troubles which would be hard to debug from an end user pov. I would say to stick on the current implementation and don't provide such magic.

It could be interesting to see how similar packages behave on this point though

ogizanagi:
> It could be interesting to see how similar packages behave on this point though

For the record, [myclabs/php-enum](https://github.com/myclabs/php-enum/blob/master/src/Enum.php#L119-L128) uses it blindly.

ogizanagi:
Closing this as I think it'll brings more issues than benefits.

mvrhov:
I also had a custom enum implementation and have used the reflection to get the constants.
[...]

PR #32, opened 2017-06-23
Add an AutoDiscoveredValuesTrait to automatically guess values from constants

ogizanagi:
Relates to #5.

A trait is a good tradeoff to me.

About naming:

1. `AutoDiscoveredValuesTrait`
1. `AutoDiscoverableValuesTrait`

?


chalasr:
Good compromise indeed! The name could probably be better, but I fail I finding one. Perhaps something with `Enumerator`. Anyway it doesn't really matter here, the current one is pretty clear. 👍  for AutoDiscoveredValues.

PR #33, opened 2017-06-27
Fix Nelmio/Alice tag for Faker provider

xavier-rdo:


ogizanagi:
Thank you @xavier-rdo

PR #35, opened 2017-06-30
AutoDiscoveredValuesTrait: do only autodiscover public constants

### LINKED ISSUE #34: Do not autodiscover private constants
[AutoDiscoveredValuesTrait](https://github.com/Elao/PhpEnums/blob/master/src/AutoDiscoveredValuesTrait.php) should not try to autodiscover private (or protected) constants.

---

https://wiki.php.net/rfc/class_const_visibility

ogizanagi:
Fixes #34

### DIFF
--- a/src/AutoDiscoveredValuesTrait.php
+++ b/src/AutoDiscoveredValuesTrait.php
@@ -25,11 +25,18 @@ trait AutoDiscoveredValuesTrait
         $enumType = static::class;
 
         if (!isset(self::$guessedValues[$enumType])) {
-            $values = (new \ReflectionClass($enumType))->getConstants();
+            $r = new \ReflectionClass($enumType);
+            $values = $r->getConstants();
+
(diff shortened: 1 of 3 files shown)


src/AutoDiscoveredValuesTrait.php
```diff
@@ -25,7 +25,14 @@ public static function values(): array
         $enumType = static::class;
 
         if (!isset(self::$guessedValues[$enumType])) {
-            $values = (new \ReflectionClass($enumType))->getConstants();
+            $r = new \ReflectionClass($enumType);
+            $values = $r->getConstants();
+
+            if (PHP_VERSION_ID >= 70100) {
+                $values = array_filter($values, function ($k) use ($r) {
```
chalasr: perhaps `string $k`?

ogizanagi: Thanks. That made me think I should filter non-integer values too for flagged enums.

PR #38, opened 2017-10-19
Update symfony 4 compatibility

benji07:
TODO:

- [x] Update tests to use new PHPUnit namespaces
- [x] Fix phpunit version used through env var
- [x] Inject a `NullLogger` (see https://github.com/symfony/symfony/pull/24385)


ogizanagi:
Thanks Benjamin.

Issue #26, opened 2017-03-08
Serializable support


mvrhov:
Is there a plan to add serializable interface to abstract class?
This then means that the enum can be serialized to a session or somewhere else.

ogizanagi:
Indeed, this feature totally makes sense to me.
If you want to, don't hesitate to provide a PR. ;)

ogizanagi:
Closing for now. See https://github.com/Elao/PhpEnums/pull/39#issuecomment-381692520

PR #39, opened 2017-10-27
Make Enum class implements \Serializable interface

skwi:
Implementation of feature asked in issue #26

Only values of type `string` and `int` are serializable with this PR since they are the recommended type in the documentation (https://github.com/skwi/PhpEnums#usage)

This is quite a constraint so let me know if it should be avoided.

### DIFF
--- a/src/Enum.php
+++ b/src/Enum.php
@@ -12,8 +12,9 @@ namespace Elao\Enum;
 
 use Elao\Enum\Exception\InvalidValueException;
 use Elao\Enum\Exception\LogicException;
+use Elao\Enum\Exception\NotSerializableException;
 
-abstract class Enum implements EnumInterface
+abstract class Enum implements EnumInterface, \Serializable
 {
     /**
      * Cached array of enum instances by enum type (FQCN).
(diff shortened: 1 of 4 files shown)


ogizanagi:
Hi @skwi ! Thanks for the PR.

Despite I approved the related issue, I've never implemented it myself because I had two concerns afterwards. I'm glad there is someone to open the discussion here:
1. The package features a really simple way to [compare enums](https://github.com/Elao/PhpEnums#compare) as each enumerated value will always return the same instance...but this does not work with deserialization, so it's flawed already. I don't know any way to workaround this, even by controlling the serialization. At least this limitation must appear in the documentation if we promote a feature about serializing an enum.
2. I never really had to care about serializing enum types using php natives but `serialize`/`unserialize` just work already fine to me without any need to get more control over it, isn't it? What would be the real benefit of using the interface? The enum value is the only thing that is part of the state, and it's unlikely (and unadvised) to have any other instance property due to the nature of enum types. So, no perf consideration is involved, right?
    
    However, we could benefit from the `\JsonSerializable` interface as an hint for integrations, like in logs with the Monolog [`NormalizerFormatter`](https://github.com/Seldaek/monolog/blob/7b992836275e09ed63c63fe33ca9993e515e6c5d/src/Monolog/Formatter/NormalizerFormatter.php#L106).

> Only values of type string and int are serializable with this PR since they are the recommended type in the documentation (https://github.com/skwi/PhpEnums#usage)

Indeed, I even replaced explicitly the docblocks from `mixed` to `int|string` because it should be the only types used for an enumerated value. Actually, the enumerated value shouldn't matter in any way and you're not supposed to rely on: only the identifier is important. As in other languages with native enum types, I'd like to generate by default internal values so you just have to declare identifiers, but that's not really doable unfortunately. I recommend strings so raw values are still human readable, and integers are mainly here for flagged enums needs.

That's why I personally think mathematic constants aren't a good use case for an enumerated type: it's not really a finite set, just an arbitrary choice list of constants in your app. But if it's really something that make sense as an enum in your app, you may just add a method returning the corresponding math constant for an instance/identifier.

PR #43, opened 2017-12-15
Use native sql type

nervo:
From @benji07 :)


ogizanagi:
Thanks @nervo! Would you also provide a feature to optionally enable this on demand? Anyway mentionning this is already great.

Issue #46, opened 2018-04-07
Design issue: Unnnecessary split of values() and readables()


ostrolucky:
```php
class myEnum {
  function values() {
    return ['value1', 'value2'];
  }

  function readables() {
    return ['value1' => 'label1', 'value2' => 'label2'];
  }
}
```
Do you see where is the problem? Why not just do
```php
class myEnum {
  function values() {
    return ['label1' => 'value1', 'label2' => 'value2'];
  }
}
```

In our case, almost all of the enums are designed to be read by FormTypes and this unnecessarily complicates the work. Of course, I solved it by proceeding to use this proposed alternative format anyway and created own basetype which doest just this.

```php
abstract class AbstractEnum extends ReadableEnum
{
    function readables()
    {
        return array_flip(static::values());
    }
}
```

But, this is just a workaround for design issue in this library, hence won't look good in eyes of other developers who are not so keen of OSS libs when they can create their own (reinvent the wheel). I'm pushing for more involvement in OSS. I spent way too much effort on searching for enum library capable enough to replace our internal implementation [everlutionsk/enum-bundle](https://github.com/everlutionsk/enum-bundle). During this process, I got burned by several others. elao/phpenums seem closest. So, let's improve it?

To figure out how to proceed, I first need to know your view on this issue. I also need to know a reason for this split. Is there some good reason for this separation?

ogizanagi:
Glad to have you here!

Actually, the reason is simple: not all enums need/are meant to be readable. What you've spotted as a design flaw would to others be interfaces segregation. Readable enums are a separated feature and the `EnumInterface::values()` method was never meant to return anything else than values. Indexes are not relevant here and despite PHP doesn't allow yet to document/typehint this properly, the actual return value of this method _should_ (but not enforced) be a set, not a dict.

However, I admit this is a DX issue, as you have to keep synchronized two methods. So I'm all in to provide a solution for this. But I won't rely on `values`. I'm for keeping clear separation for the most "low-level" classes and allow to compose higher-level, more opinionated classes. The `AutoDiscoveredValuesTrait` was already a step in this way and I'm happy about it.

So, [here is my suggestion](https://github.com/Elao/PhpEnums/pull/49) for this issue. I hope you'll be satisfied this way :)

ostrolucky:
[...]

PR #47, opened 2018-04-07
Change AutoDiscoveredValuesTrait to use constant values as keys

ostrolucky:
Hi, can I have this change? It aids with https://github.com/Elao/PhpEnums/issues/46

For some enums, it's ok to have label same as value. In those cases, this allows to not having to define neither values() and neither readables(). Shouldn't pose too big risk of BC break.

### DIFF
--- a/src/AutoDiscoveredValuesTrait.php
+++ b/src/AutoDiscoveredValuesTrait.php
@@ -40,7 +40,7 @@ trait AutoDiscoveredValuesTrait
                 });
             }
 
-            self::$guessedValues[$enumType] = array_values($values);
+            self::$guessedValues[$enumType] = array_combine($values, $values);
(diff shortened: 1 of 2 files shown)


ogizanagi:
This would be solved by https://github.com/Elao/PhpEnums/pull/49 with no potential BC issue, in respect of the initial `EnumInterface::values` contract.

However, I'm asking you: would you prefer default labels being the value itself, or a human representation of the constant name [as done in default implementation of the FlaggedEnum](https://github.com/Elao/PhpEnums/blob/master/tests/Unit/FlaggedEnumTest.php#L136-L153)?

ostrolucky:
Thank you for welcoming similar idea. Let's continue in #49

PR #48, opened 2018-04-07
Add ScalarToEnumTransformer

ostrolucky:
I found EnumType very limiting, because it enforces either Enum instances inside choices, or scalar value as output.

This is inconvenient, because it's common to have scalar key-value pairs stored as an array in class constant. Storing there objects is not possible.

So, this transformer is meant to be used in conjunction with normal ChoiceType and scalar choices.


ostrolucky:
Of course. Just didn't want to spend more time on figuring out how to integrate this there. It can be used standalone for now. We can integrate it later :)

ogizanagi:
Thanks @ostrolucky.

PR #50, opened 2018-04-08
Allow usage of FCQN instead of aliases in Faker provider

ostrolucky:
This allows avoiding defining alias map if I want to use this provider


tests/Unit/Bridge/Faker/Provider/EnumProviderTest.php
```diff
@@ -40,6 +45,9 @@ public function testRandomEnumMethodShouldReturnExpectedEnums()
         $simple = $enumProvider->randomEnum('Simple');
         $this->assertInstanceOf(SimpleEnum::class, $simple);
 
+        $gender = $enumProvider->enum('Elao\Enum\Tests\Fixtures\Enum\Gender::MALE');
```
ogizanagi: It's supposed to test `randomEnum` here 😄

ogizanagi:
Thank you @ostrolucky.

PR #58, opened 2018-04-28
php-translation extractor

ostrolucky:


ogizanagi:
Thanks @ostrolucky for taking care of this.
I've got some questions though. How would you use it exactly? (With and without the Symfony bundle for instance)
I don't know much about the extractors, but reading the documentation, I wonder if we couldn't provide a better integration by relying on the `PHPFileExtractor` and providing a visitor as explained [here](http://php-translation.readthedocs.io/en/latest/guides/adding-extractor.html) and considering the described architecture [here](http://php-translation.readthedocs.io/en/latest/components/extractor.html)? It's a lot more work AFAIU, but seems to be worth it, right?

ostrolucky:
That was my first idea. There already is very similar extractor built in there, called sourcelocationcontainervisitor. But that approach is a mistake imho, because it's too stupid. In user space I would have to make each enum implement interface separately. See https://github.com/php-translation/extractor/issues/81#issuecomment-385190697

Usage is to have a class where extractors are injected. Calling it, it visits all of them and spits out all translations

ogizanagi:
> There already is very similar extractor built in there, called sourcelocationcontainervisitor. But that approach is a mistake imho, because it's too stupid. In user space I would have to make each enum implement interface separately. It's also a lot more complicated and slower.

`TranslationSourceLocationContainer` seems to exist for most dynamic needs. I don't think either we should use it indeed. Implementing interfaces on each enum type is a no-go. But by having a whole new visitor able to detect all classes implementing `ReadableEnumInterface` in `src`, no config and seamless integration with the existing code, we could make it, right? Obviously it'll be slower looking at all application sources, but is it actually an issue here?

ostrolucky:
We don't use a bundle, just standalone extractor component. But I doubt it can be used easily in a bundle. Usage will be something like

```php
class TranslationGenerator {
  public function __construct(Finder $finder, Extractor $officialExtractor, ReflectionEnumExtractor $enumExtractor)
  {
      $this->finder = $finder;
      $this->extractor = $officialExtractor;
      $this->enumExtractor = $enumExtractor;
  }
  public function generate(): SourceCollection {
    return $this->enumExtractor->getSourceLocations($this->extractor->extract($this->finder));
  }
}
```

but I don't feel like advertising that in readme.

> but by having a whole new visitor able to detect all classes implementing ReadableEnumInterfac

That would have same limitations SourceLocationContainerVisitor have. It wouldn't be able to collect enums which don't implement ReadableEnumInterface directly, but do under several layers.

ostrolucky:
Ok I will write visitor.

ostrolucky:
@klinec this is the PR I was talking about. Feel free to grab the code from it.

I don't need this as much to justify effort of rewriting it into visitor.

PR #59, opened 2018-04-29
Add EnumValueResolver

ostrolucky:


ostrolucky:
IMHO it's an edge case and 99% developers will rather rename variable in case of conflict than change wiring of this resolver. I can change the order if you wish though.

tests/Integration/Bridge/Symfony/ArgumentResolverTest.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+
+class ArgumentResolverTest extends WebTestCase
+{
+    public function testValidEnumPassedInRequest()
+    {
+        if (!interface_exists(ArgumentValueResolverInterface::class)) {
+            $this->markTestSkipped(sprintf('"%s" is not present int this Symfony version', ArgumentValueResolverInterface::class));
```
ostrolucky: typo

ogizanagi:
Oops. Sorry for the lack of response. 
Let's go as is. We cannot assume if the `GET` vs `POST` override is legit or not, so no exception.

ostrolucky:
Good. I've fixed the typo

src/Bridge/Symfony/HttpKernel/Controller/ArgumentResolver/EnumValueResolver.php
```diff
@@ -0,0 +1,44 @@
...
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+
+class EnumValueResolver implements ArgumentValueResolverInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(Request $request, ArgumentMetadata $argument)
+    {
+        return is_subclass_of($argument->getType(), EnumInterface::class);
```
ogizanagi: `!$argument->isVariadic()` if we don't want to support variadics. Otherwise the resolve method should also handle arrays.

We must also check if the `$argument->getName()` is part of the attributes, query or request bags (using `has` on each).

src/Bridge/Symfony/HttpKernel/Controller/ArgumentResolver/EnumValueResolver.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function resolve(Request $request, ArgumentMetadata $argument)
+    {
+        /** @var EnumInterface $enumClass */
+        $enumClass = $argument->getType();
+        $value = $request->get($argument->getName());
+
+        if (!$enumClass::accepts($value)) {
```
ogizanagi: `$value` could be `null` and expected if `$argument->isNullable()`, right?

src/Bridge/Symfony/HttpKernel/Controller/ArgumentResolver/EnumValueResolver.php
ostrolucky: What do you mean? Why do we need those has checks?

ogizanagi: See `RequestAttributeValueResolver` for instance. The support method should return `true` only if the argument can be found in the request. If it isn't, the default value provided on the controller will be used (if provided, otherwise an exception).

README.md
```diff
@@ -537,6 +538,19 @@ abstract class AbstractEnumType extends Type
...
+<a href="https://symfony.com"><img src="https://img.shields.io/badge/Symfony-3.1%2B-green.svg?style=flat-square" title="Available for Symfony 3.1+" alt="Available for Symfony 3.1+" align="right"></a>
+
+There is an [argument value resolver](https://symfony.com/doc/current/controller/argument_value_resolver.html) provided,
+which seamlessly transforms HTTP request parameter into enum instance, just by typehinting the target enum in controller action.
+
+Simply register it inside the DIC configuration:
+```yaml
+services:
+    Elao\Enum\Bridge\Symfony\HttpKernel\Controller\ArgumentResolver\EnumValueResolver:
+        tags: [{ name: controller.argument_value_resolver }]
```
ogizanagi: Shouldn't the priority be set higher than 100 in order to be executed before `RequestAttributeValueResolver`, so you can still get a request attribute value as enum instead of raw value?

ostrolucky:
failures I believe unrelated

composer.json
```diff
@@ -28,14 +28,14 @@
     "prefer-stable": true,
     "minimum-stability": "dev",
     "require": {
-        "php": ">=7.0"
+        "php": ">=7.0",
+        "symfony/symfony": "^2.8"
```
chalasr: should be reverted, this package can be used without any symfony component

src/Bridge/Symfony/HttpKernel/Controller/ArgumentResolver/EnumValueResolver.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function resolve(Request $request, ArgumentMetadata $argument)
+    {
+        /** @var EnumInterface $enumClass */
+        $enumClass = $argument->getType();
+        $requestValue = $request->get($argument->getName());
```
ogizanagi: Could just use:

```diff
- $requestValue = $request->get($argument->getName());
+ $requestValue = (array) $request->get($argument->getName());
```

and remove further if

ostrolucky:
Let me do one more change

ostrolucky:
ready

ogizanagi:
@chalasr : Good to you too ? :)

ogizanagi:
Thank you @ostrolucky.

PR #60, opened 2018-05-06
Fix return type in ValueToEnumTransformer::transform

ostrolucky:


ogizanagi:
Good catch, thanks @ostrolucky.

PR #64, opened 2018-07-27
Fix Enum constraint deserialization

ogizanagi:
When the constraint is serialized (to be cached) and the `choice` option is provided (and expecting enum instances and not just values), deserialization of the choice option will not use `EnumInterface::get()` and thus won't create the proper multiton instances, making strict comparisons fail.
This fixes it by replacing deserialized instances by proper ones.

### DIFF
--- a/src/Bridge/Symfony/Validator/Constraint/Enum.php
+++ b/src/Bridge/Symfony/Validator/Constraint/Enum.php
@@ -95,4 +95,22 @@ class Enum extends Choice
     {
         return ['class'];
     }
+
+    /**
+     * Fixup deserialized enum instances by replacing them with actual multiton instances,
+     * so strict comparison works.
+     *
+     * {@inheritdoc}
+     */
+    public function __wakeup()
+    {
+        if (!$this->asValue && \is_array($this->choices)) {
+            $this->choices = array_map(function (EnumInterface $enum): EnumInterface {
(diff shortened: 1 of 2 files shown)


src/Bridge/Symfony/Validator/Constraint/Enum.php
```diff
@@ -95,4 +95,22 @@ public function getRequiredOptions()
...
+     * Fixup deserialized enum instances by replacing them with actual multiton instances,
+     * so strict comparison works.
+     *
+     * {@inheritdoc}
+     */
+    public function __wakeup()
+    {
+        if (!$this->asValue && \is_array($this->choices)) {
+            $this->choices = array_map(function (EnumInterface $enum) {
+                /** @var string|EnumInterface $enumClass */
```
chalasr: wrong?

ogizanagi: That's only to provide autocompletion. used in multiple places already.

chalasr: I mean the docblcok position, `get_class` returns a string, shouldn't it be under the return below?

chalasr: oh ok got it, strange :)

PR #67, opened 2018-08-13
Use `defined` call to check if constant exists

pierredup:
Use `defined` to check if a constant is defined on a class instead of trying to get the constant value and suppressing errors (which has a slightly negative impact on performance)


ogizanagi:
Thank you @pierredup.

PR #71, opened 2019-06-28
Some fixes to be more SF4 compliant

alanpoulain:
This is a great bundle! I would like to see it more used.


ogizanagi:
Thanks Alan! I like it!

PR #73, opened 2019-08-09
Introduce readablesFor method

maximecolin:
```php
MyEnum::readablesFor(['FOOBAR', 'BARFOO']);
```

### DIFF
--- a/src/ReadableEnumInterface.php
+++ b/src/ReadableEnumInterface.php
@@ -33,6 +33,17 @@ interface ReadableEnumInterface extends EnumInterface
     public static function readableFor($value): string;
 
     /**
+     * Gets the human representations for given values.
+     *
+     * @param array $values
+     *
+     * @throws InvalidValueException When $value is not acceptable for this enumeration type
+     *
+     * @return array
+     */
+    public static function readablesFor(array $values): array;
+
+    /**
      * Gets the human representation of the value.
      *
      * @return string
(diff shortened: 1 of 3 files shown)


ogizanagi:
Hmm, I'm not sure it is very common nor worth it compared to: 

```php
array_map([MyEnum::class, 'readableFor'], ['Male', 'Female']);
```

What do you think?

maximecolin:
Well in my case it makes my code much readable and simpler.

src/ReadableEnumInterface.php
```diff
@@ -32,6 +32,17 @@ public static function readables(): array;
...
     public static function readableFor($value): string;
 
+    /**
+     * Gets the human representations for given values.
+     *
+     * @param array $values
+     *
+     * @throws InvalidValueException When $value is not acceptable for this enumeration type
+     *
+     * @return array
```
ogizanagi: `string[]`?

src/ReadableEnumInterface.php
```diff
@@ -32,6 +32,17 @@ public static function readables(): array;
...
+     */
+    public static function readablesFor(array $values): array;
```
ogizanagi: Adding it to the interface if a BC break.
But Symfony's debug class loader allows to hint unimplemented `@methods` since https://github.com/symfony/symfony/pull/28902. So it could be a solution.

ogizanagi:
I'm closing as I don't think it's worth it. Such helpers shouldn't be part of the interface anyway.
I might reconsider something similar later in an `EnumSet` class with more utility methods to manipulate and format enums, but not in the main interface.

PR #82, opened 2020-04-16
Minor translation extractor config enhancements

ogizanagi:
Just some minor improvements after #81.

@danut007ro: If you have some time to review those changes, I'll appreciate!


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -76,15 +76,9 @@ public function getConfigTreeBuilder()
...
-                        ->arrayPrototype()
-                            ->beforeNormalization()
-                                ->ifString()->then(static function (string $v): array { return ['namespace' => $v]; })
-                            ->end()
-                            ->children()
-                                ->scalarNode('namespace')->isRequired()->cannotBeEmpty()->end()
-                            ->end()
+                        ->example(['App\Enum' => '%kernel.project_dir%/src/Enum'])
+                        ->useAttributeAsKey('namespace')
+                        ->scalarPrototype()
```
danut007ro, line 81: Won't it be better to have this as array, just like doctrine types where you can have two properties specified (name and type). This would be a safeguard for future updates to add more properties to a mapping (eg. have another `domain` for a namespace)

ogizanagi: That's wouldn't be much an issue actually using normalization for BC compat if we want to support such cases later.

danut007ro: Then I'm ok with this. Thank you

PR #84, opened 2020-04-23
add ApiPlatform ElaoEnum schema type

WhiteRabbitDE:
This PR adds a class that decorates the ApiPlatform TypeFactory to support the documentation of elao/PhpEnums when generating the swagger ui.

Before :
enum-before

After:
enum-after

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ Table of Contents
     * [Symfony VarDumper component](#symfony-vardumper-component)
     * [Faker](#faker)
       * [Usage with Alice](#usage-with-alice)
+    * [Api-Platform](#api-platform)
+      * [OpenApi / Swagger](#OpenApi-/-Swagger)
   * [API](#api)
     * [Simple enum](#simple-enum)
     * [Readable enum](#readable-enum)
@@ -865,6 +867,11 @@ MyEntity:
 ```
 
 > 📝 `MISTER` in `<enum(Civility::MISTER)>` refers to a constant defined in the `Civility` enum class, not to a constant's value ('mister' string for instance).
+
+## Api-Platform
+
+### OpenApi / Swagger
+The PhpEnums library provides an `Elao\Enum\Bridge\ApiPlatform\Core\JsonSchema\Type\ElaoEnumType` to generate a OpenApi (formally Swagger) documentation based on your enums. This decorator is automatically wired for you when using this library. 
 
 # API
 
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "php": ">=7.0"
     },
     "require-dev": {
+        "api-platform/core": "^2.5.1",
         "doctrine/data-fixtures": "^1.2",
         "doctrine/doctrine-bundle": "^1.4|^2.0",
         "doctrine/orm": "^2.4",
--- a/src/Bridge/ApiPlatform/Core/JsonSchema/Type/ElaoEnumType.php
+++ b/src/Bridge/ApiPlatform/Core/JsonSchema/Type/ElaoEnumType.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of the "elao/enum" package.
+ *
+ * Copyright (C) Elao
+ *
+ * @author Elao <contact@elao.com>
+ */
+
+namespace Elao\Enum\Bridge\ApiPlatform\Core\JsonSchema\Type;
+
+use ApiPlatform\Core\JsonSchema\Schema;
+use ApiPlatform\Core\JsonSchema\TypeFactoryInterface;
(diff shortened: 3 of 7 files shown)


WhiteRabbitDE:
Im not sure what to do about the test. @ogizanagi can you help me out here ?

ogizanagi:
For 7.0 build, [let's drop the package from `.travis.yaml`](https://github.com/Elao/PhpEnums/blob/f4f9c9190b66a1aa25900a9ab8c893fb3ecefdd3/.travis.yml#L62) (we could probably drop php 7.0 support as it's end of life, but maybe not just for this) and mark the tests with `@requires PHP 7.1`.

For Sf 5.x build, you should replace `"api-platform/core": "2.5"` by `"api-platform/core": "^2.5.1"`

README.md
```diff
@@ -866,6 +868,29 @@ MyEntity:
...
+### OpenApi / Swagger
+The PhpEnums library provides an `Elao\Enum\Bridge\ApiPlatform\Core\JsonSchema\Type\ElaoEnumType` to generate a OpenApi (formally Swagger) documentation based on your enums.
+
+To use it, you need to decorate the `api_platform.json_schema.type_factory` service.
+
+```yml
+# config/services.yaml
+services:
+    # symfony <= 5.0
+    Elao\Enum\Bridge\ApiPlatform\Core\JsonSchema\Type\ElaoEnumType:
```
ogizanagi, line 882: Let's do this directly from the bundle extension if the decorated factory exists, wdyt?
You could write this in the [`services.xml`](https://github.com/Elao/PhpEnums/blob/master/src/Bridge/Symfony/Bundle/Resources/config/services.xml) and drop the definition if the class doesn't exist.

Is there any drawback/potential harm registering this service automatically without the ability to disable it?

composer.json
```diff
@@ -31,9 +31,10 @@
     "prefer-stable": true,
     "minimum-stability": "dev",
     "require": {
-        "php": ">=7.0"
+        "php": ">=7.1"
     },
     "require-dev": {
+        "api-platform/core": "2.5",
```
ogizanagi, line 37:
```suggestion
        "api-platform/core": "^2.5.1",
```

src/Bridge/ApiPlatform/Core/JsonSchema/Type/ElaoEnumType.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
```
ogizanagi, line 3: We're not using strict types yet in the code base. Let's remove it for now.

README.md
WhiteRabbitDE: I dont think so. The only case would be if you want to write your own decorator for this lib which is unlikely imho.

I will add it to the services.xml.

composer.json
WhiteRabbitDE: Good catch ! I thought I added it with patch level... my bad.

src/Bridge/ApiPlatform/Core/JsonSchema/Type/ElaoEnumType.php
WhiteRabbitDE: Removed.

WhiteRabbitDE:
> For 7.0 build, [let's drop the package from `.travis.yaml`](https://github.com/Elao/PhpEnums/blob/f4f9c9190b66a1aa25900a9ab8c893fb3ecefdd3/.travis.yml#L62) (we could probably drop php 7.0 support as it's end of life, but maybe not just for this) and mark the tests with `@requires PHP 7.1`.
> 
That change would be part of another PR, right ?

ogizanagi:
> > For 7.0 build, [let's drop the package from `.travis.yaml`](https://github.com/Elao/PhpEnums/blob/f4f9c9190b66a1aa25900a9ab8c893fb3ecefdd3/.travis.yml#L62) (we could probably drop php 7.0 support as it's end of life, but maybe not just for this) and mark the tests with `@requires PHP 7.1`.
> 
> That change would be part of another PR, right ?

No you should add it in this PR to make travis happy :)
Just add a check removing the package in the travis file according to the php version.
Only mark the api platform `ElaoEnumTypeTest` class with `@requires PHP 7.1`, and that should do it!

README.md
ogizanagi: Nice, thank you.

You can now remove this wiring from the documentation (you can keep the section & first sentence, but mention everything should be wired for you when using the bundle)

WhiteRabbitDE:
PR is a mess but at least the test are successful now :P

ogizanagi:
Thanks @WhiteRabbitDE for working on this feature, this is much appreciated.

PR #85, opened 2020-04-24
Update readme for symfony translation extractor integration

danut007ro:


ogizanagi:
Good catch, thanks @danut007ro.

PR #87, opened 2020-05-07
[Doctrine] Allow to configure ENUM(...values) SQL declaration

### LINKED ISSUE #86: TypesDumper - Extends
Hi !

I want to know if it will be possible to extends the TypesDumper class to override getSQLDeclaration method ?

Thanks

https://github.com/Elao/PhpEnums/blob/3afa22c65bad1822c2c535c870f8f45c76509e43/src/Bridge/Symfony/Bundle/DependencyInjection/Compiler/DoctrineDBALTypesPass.php#L35

ogizanagi:
Fixes #86 

@yvalentin: What do you think about this strategy? I'd appreciate if you could give it a try in a real-world project and give me feedbacks 🙏

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,15 +481,27 @@ This configuration is equivalent to the following sections explaining how to cre
 elao_enum:
     doctrine:
         types:
-            App\Enum\GenderEnum: gender
-            App\Enum\Permissions: { name: permissions, type: int } # values are stored as integers
+            App\Enum\GenderEnum: gender # Defaults to `{ name: gender, type: string }` for string based enum (translates to VARCHAR)
+            App\Enum\AnotherEnum: { name: another, type: enum } # string based enum with SQL ENUM column definition
+            App\Enum\Permissions: { name: permissions, type: int } # values are stored as integers. Default for flagged enums.
 ```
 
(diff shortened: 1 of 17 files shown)


ogizanagi:
@benji07 : Yes indeed. That's one of the reasons (appart for BC reasons) I won't enable this by default in the bundle. But anyone willing to use it despite this drawback can now benefit from this and handle the migration themselves 🤷 . 
Thx for the review.

yvalentin:
@ogizanagi @benji07 I kown this problem about updating the enum definition.

I can test your modification on the current project I'm working on.
And I will give you a feelback soon.

Thank for this !

README.md
```diff
@@ -481,15 +481,27 @@ This configuration is equivalent to the following sections explaining how to cre
 elao_enum:
     doctrine:
         types:
-            App\Enum\GenderEnum: gender
-            App\Enum\Permissions: { name: permissions, type: int } # values are stored as integers
+            App\Enum\GenderEnum: gender # string based enum (translates to VARCHAR)
```
yvalentin, line 484: May be here, you can add the type key on comment for VARCHAR => (type: string)

Issue #93, opened 2020-08-27
make it easier to create enum from "dumb" enums from other libraries


mathroc:
a lot of libraries have enum classes that are only contants bags (eg: https://github.com/Mangopay/mangopay2-php-sdk/blob/master/MangoPay/EventType.php or https://github.com/symfony/http-kernel/blob/master/KernelEvents.php)

I’d like to tranform those into `Elao\Enum\EnumInterface` but I can’t extend both the 3rd party class and `Elao\Enum\Enum` so I have to reimplement `Elao\Enum\EnumInterface` (by copying, the content of `Elao\Enum\Enum`).

It’d be nice if this was available as a trait:

```
class EventType extends \MangoPay\EventType implements EnumInterface
{
    use EnumTrait;
    use AutoDiscoveredValuesTrait;
}
```

ogizanagi:
I don't think relying on extending a 3rd party class "enum" is a good idea for this.
We could perhaps propose such a thing:

```php
class EventType extends Enum
{
    private const INHERIT_CONSTS_FROM = \MangoPay\EventType::class;

    use AutoDiscoveredValuesTrait;
}
```

where the `AutoDiscoveredValuesTrait` would declare values from the public consts in `\MangoPay\EventType`,and then instantiating such an enum would be like:

```php
EventType::get(\MangoPay\EventType:: KycCreated);
```

or we could imagine a command generating a proper enum class from a "dumb enum".

in practice, I did create such enums wrapping another third-party "dumb enum", but it was also expliciting the one values I cared about in my application, not just blindly inherit all.
So in the end, I'm just not sure if it deserves a new feature.

mathroc:
> I don't think relying on extending a 3rd party "enum" class is a good idea for this.
> We could perhaps propose such a thing:
> 
> ```
> class EventType extends Enum
> {
>     private const INHERIT_CONSTS_FROM = \MangoPay\EventType::class;
> 
>     use AutoDiscoveredValuesTrait;
> }
> ```
> 
> where the `AutoDiscoveredValuesTrait` would declare values from the public consts in `\MangoPay\EventType`. Then instantiating such an enum would be like:
> 
> ```
> EventType::get(\MangoPay\EventType:: KycCreated);
> ```

I don’t mind either way. It just seems more complicated with `INHERIT_CONSTS_FROM` to me and I don’t get what advantages it has

> 
> or we could imagine a command generating a proper enum class from a "dumb enum".
> 
[...]

PR #94, opened 2020-09-01
make it easier to create enum from "dumb" enums

mathroc:
This is an implementation of the idea in #93 if you think it’s worth it

It’s very straightforward, everything in Enum was just moved into EnumTrait and EnumTrait is now "use"-ed in Enum to avoid any BC breaks


ogizanagi:
Closed in favor of #96

PR #96, opened 2020-09-03
Allow to discover values from other classes

### LINKED ISSUE #93: make it easier to create enum from "dumb" enums from other libraries

ogizanagi:
Fixes #93 
Supersedes #94 

cc @mathroc

### Todo

- [x] Fix caster

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,17 @@ final class Gender extends Enum
 }
 ```
 
+Get an instance of your enum type:
+
+```php
+<?php
+$enum = Gender::get(Gender::Male);
+```
+
+You can easily retrieve the enumeration's value by using `$enum->getValue();`
+
+> 📝 Enum values are supposed to be integers or strings.
+
 > 📝 It's recommended to make your enums classes `final`, because it won't make sense to extend them in most situations (unless you're creating a new base enum type), and you won't need to mock an enum type.
 
 > 💡 You can also use the `AutoDiscoveredValuesTrait` to automagically guess values from the constants defined in your enum, so you don't have to implement `EnumInterface::values()` yourself:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -181,16 +192,45 @@ final class Gender extends Enum
...
+    use AutoDiscoveredValuesTrait;
+
+    protected static function getDiscoveredClasses(): array
+    {
+        return [self::class, MangoPay\EventType::class];
+    }
+}
+
+# Usage:
+MangoPayEventType::get(MangoPay\EventType::KycCreated);
```
mathroc, line 232: does `MangoPayEventType::KycCreated()` still work?

tests/Unit/AutoDiscoveredValuesTraitTest.php
```diff
@@ -44,6 +45,11 @@ public function testThrowsOnChoicesMisuses()
 
         AutoDiscoveredEnumMisusingChoices::foo();
     }
+
+    public function testItAutoDiscoversValuesFromDumbEnum()
+    {
+        $this->assertSame(['foo', 'bar', 'baz'], AutoDiscoveredDumbEnum::values());
```
mathroc, line 51: I don’t really mind but since the discovered classes are `[self::class, DumbEnum::class]` I would have expected `'bar'` to be first

mathroc (approved):
looks nice !

README.md
ogizanagi: No, as the contract for this magic calls to work is to have a matching constant in the enum class.
It could be improved though.

PR #98, opened 2020-10-20
Allow user to create collection of enum with doctrine

benji07:
- [x] Documentation
- [x] Mark legacy format of configuration as deprecated

### DIFF
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -55,7 +55,7 @@ class Configuration implements ConfigurationInterface
                             foreach ($values as $name => $config) {
                                 if (class_exists($name)) {
                                     $legacyFormat = true;
-                                    // need to flip
+                                    @trigger_error('Using enum FQCN as keys at path "elao_enum.doctrine.types" is deprecated. Provide the name as keys and add the "class" option for each entry instead.', E_USER_DEPRECATED);
                                     break;
                                 }
                             }
--- a/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
@@ -49,24 +49,54 @@ class Configuration implements ConfigurationInterface
                     ->info('If true, by default for string enumerations, generate DBAL types with an ENUM SQL declaration with enum values instead of a VARCHAR (Your platform must support it)')
                 ->end()
                 ->arrayNode('types')
-                    ->validate()
-                        ->ifTrue(static function (array $v): bool {return self::hasNonEnumKeys($v); })
-                        ->then(static function (array $v) { self::throwsNonEnumKeysException($v); })
+                    ->beforeNormalization()
+                        ->always(function ($values) {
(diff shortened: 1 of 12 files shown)


src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -49,24 +49,53 @@ private function addDoctrineSection(ArrayNodeDefinition $rootNode)
...
+                            } else {
+                                $newValues = $values;
+                            }
+
+                            return $newValues;
+                        })
+                    ->end()
                     ->validate()
                         ->ifTrue(static function (array $v): bool {return self::hasNonEnumKeys($v); })
                         ->then(static function (array $v) { self::throwsNonEnumKeysException($v); })
```
ogizanagi, line 81:
```suggestion
                        ->ifTrue(static function (array $v): bool {return self::hasNonEnumClasses($v); })
                        ->then(static function (array $v) { self::throwsNonEnumClassesException($v); })
```

ogizanagi: Can't we add this validation on the `class` node directly now it's not the keys anymore?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -55,7 +55,7 @@ private function addDoctrineSection(ArrayNodeDefinition $rootNode)
                             foreach ($values as $name => $config) {
                                 if (class_exists($name)) {
                                     $legacyFormat = true;
-                                    // need to flip
+                                    @trigger_error('The format has change at path elao_enum.doctrine.types, you must index your configuration array by name and not by class fqcn', E_USER_DEPRECATED);
```
ogizanagi, line 58:
```suggestion
                                    @trigger_error('Using enum FQCN as keys at path "elao_enum.doctrine.types" is deprecated. provide the name as keys and add the "class" option for each entry instead.', E_USER_DEPRECATED);
```

tests/Fixtures/Bridge/Symfony/Bundle/DependencyInjection/ElaoEnumExtension/xml/doctrine_types_legacy.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xmlns:elao-enum="http://elao.com/schema/dic/elao_enum"
+           xsi:schemaLocation="http://elao.com/schema/dic/elao_enum http://elao.com/schema/dic/elao_enum/elao_enum.xsd">
+
+    <elao-enum:config>
```
ogizanagi, line 8: No changes for the XML format, so you can simply skip it in the test case

benji07: yes, i just added the test in the wrong test case

ogizanagi:
Thanks @benji07.

PR #107, opened 2020-11-06
Fix autodiscovered values contiguous indices

### LINKED ISSUE #105: EnumProvider::randomEnum()  - sometimes null is returned
Hey,

in my fixtures (alice bundle) i use 

`mimeType: <randomEnum(MediaEnum)>`

When running my testsuite, sometimes the error below occures. This happens for other fixtures/enums too. I have this error for about 1-2 weeks now, never had this error before. Im not sure if it is caused by an update of elao/phpenums or another library i updated.

After running the same testsuite again (without chaning anything) he error is gone and the data is generated correctly. Does anyone have an idea what the cause could be ?

```
vendor/elao/enum/src/Bridge/Faker/Provider/EnumProvider.php on line 90

In SimpleObjectGenerator.php line 112:
                                                                               
  An error occurred while generating the fixture "attachment_2" (App\Entity\A  
  ttachment): Return value of Elao\Enum\Bridge\Faker\Provider\EnumProvider::r  
  andomEnum() must implement interface Elao\Enum\EnumInterface, null returned  
                                                                               

In EnumProvider.php line 90:
                                                                               
  Return value of Elao\Enum\Bridge\Faker\Provider\EnumProvider::randomEnum()   
  must implement interface Elao\Enum\EnumInterface, null returned    
```

ogizanagi:
Fixes #105 

cc @WhiteRabbitDE

### DIFF
--- a/tests/Unit/AutoDiscoveredValuesTraitTest.php
+++ b/tests/Unit/AutoDiscoveredValuesTraitTest.php
@@ -50,6 +50,11 @@ class AutoDiscoveredValuesTraitTest extends TestCase
     {
         $this->assertSame(['foo', 'bar', 'baz'], AutoDiscoveredDumbEnum::values());
     }
+
+    public function testItRemovesAutoDiscoveredNonUniqueIdentifiersWithContiguousIndices()
+    {
+        $this->assertSame([0 => 'foo', 1 => 'bar'], AutoDiscoveredNonUniqueEnumIndentifiers::values());
+    }
 }
 
 final class AutoDiscoveredEnum extends Enum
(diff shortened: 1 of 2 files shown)


tests/Unit/AutoDiscoveredValuesTraitTest.php
```diff
@@ -50,6 +50,11 @@ public function testItAutoDiscoversValuesFromDumbEnum()
     {
         $this->assertSame(['foo', 'bar', 'baz'], AutoDiscoveredDumbEnum::values());
     }
+
+    public function testItRemovesAutoDiscoveredNonUniqueIdentifiersWithContiguousIndices()
+    {
+        $this->assertSame([0 => 'foo', 1 => 'bar'], AutoDiscoveredNonUniqueEnumIndentifiers::values());
```
ogizanagi, line 56: (of course indices are unnecessary here, just making it obvious for everyone)

WhiteRabbitDE (approved):
Looks good to me. I also tested this fix in my real life application and it works like a charm.
Thx @ogizanagi  :-)

PR #109, opened 2020-11-09
Fix invalid libPath on windows when using unix path in config

### LINKED ISSUE #108: JSDumper - Invalid import path generated ( windows maschines )
Hi,

when using the JSDumper the generated import path is not valid. Test "testDumpEnumToFIle" is also failing because the expectedImportPath is e.g. "'../../lib/enum', but the acutal result is '../../lib/enum.js/'.

Result in the application 
```
//acutal
import { ReadableEnum } from '../../lib/enum.js/';
 
//expected
import { ReadableEnum } from '../../lib/enum';
```

with the following configuration
``` 
 js:
        base_dir: '%kernel.project_dir%/assets/js/symfony/enum'
        lib_path: '%kernel.project_dir%/assets/js/lib/enum.js'
        paths:
            App\Enum\ValueType: 'ValueType.js'
``` 

The reason is that the rtrim uses the DIRECTORY_SEPERATOR which is "\\" on windows and so the preg_replace does not remove the / after .js.
``` 
        $relativeLibPath = preg_replace('#.js$#', '', rtrim(
            $this->fs->makePathRelative(realpath($this->libPath), realpath(\dirname($path))),
            DIRECTORY_SEPARATOR
        ));
``` 

We could replace DIRECTORY_SEPERATOR with "\\\\/" or adjust the .js$ to .js/$. I prefer extending the characters list.

WhiteRabbitDE:
Fixes #108

### DIFF
--- a/src/JsDumper/JsDumper.php
+++ b/src/JsDumper/JsDumper.php
@@ -91,7 +91,7 @@ JS;
     {
         $relativeLibPath = preg_replace('#.js$#', '', rtrim(
             $this->fs->makePathRelative(realpath($this->libPath), realpath(\dirname($path))),
-            DIRECTORY_SEPARATOR
+            '\\/'
         ));
 
         if (is_a($enumFqcn, FlaggedEnum::class, true)) {


ogizanagi:
Thanks @WhiteRabbitDE.

Issue #110, opened 2020-11-09
Add an AppVeyor build


ogizanagi:
So we can ensure the tests passe on a Windows env

See https://github.com/symfony/symfony/blob/5.x/.appveyor.yml for an example

PR #111, opened 2020-11-09
add appveyor configuration

WhiteRabbitDE:
#110 

This is an inital appveyor config using php 7.4. Feedback is welcome.

### DIFF
--- a/src/Bridge/Symfony/Translation/Extractor/EnumExtractor.php
+++ b/src/Bridge/Symfony/Translation/Extractor/EnumExtractor.php
@@ -12,6 +12,7 @@ namespace Elao\Enum\Bridge\Symfony\Translation\Extractor;
 
 use Elao\Enum\ReadableEnum;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\Glob;
 use Symfony\Component\Finder\SplFileInfo;
 use Symfony\Component\Translation\Extractor\ExtractorInterface;
 use Symfony\Component\Translation\MessageCatalogue;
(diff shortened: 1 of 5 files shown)


ogizanagi:
I've enabled AppVeyor on the repo. It should trigger on new push.

WhiteRabbitDE:
@ogizanagi I think thats it ! There are 3 tests failing but those are also failing on my local machine so i think thats ok.

ogizanagi:
Congrats 👏 
Would you have some time to inspect those tests and attempt to fix it? 🙏

WhiteRabbitDE:
I went through the errors and fixed the test. Unfortunatly i really dislike some of the fixes... Most problems are caused by the directory seperator which is "/" in linux and "\" on windows. 
Maybe we can have a short chat on slack or smth before i push those changes.

src/Bridge/Symfony/Translation/Extractor/EnumExtractor.php
```diff
@@ -113,7 +112,7 @@ public function extract($resource, MessageCatalogue $catalog)
 
                     $catalog->set($translationKey, $this->prefix . $translationKey, $this->domain);
                     $metadata = $catalog->getMetadata($translationKey, $this->domain) ?? [];
-                    $normalizedFilename = preg_replace('{[\\\\/]+}', '/', $file->getPathName());
+                    $normalizedFilename = preg_replace('{[\\\\/]+}', DIRECTORY_SEPARATOR, $file->getPathName());
```
ogizanagi, line 115: I think this one comes from https://github.com/symfony/symfony/blob/b285584e77e69ad9740711a12581c3fce46535ee/src/Symfony/Component/Translation/Extractor/PhpExtractor.php#L295 and should stay the same :)

ogizanagi (approved):
🙏

src/Bridge/Symfony/Translation/Extractor/EnumExtractor.php
WhiteRabbitDE: Reverting this change would make the testGenerateTranslations test fail because the path we assert on __DIR__.

We could revert this and use the preg_replace in the assert too.

``` 
$this->assertEquals(
            [
                'trans_readable_enum' => ['sources' => [ preg_replace('{[\\\\/]+}', '/',__DIR__ . '/Enum/ReadableEnum.php:readable_enum')]],
            ],
            $catalog->getMetadata('', 'messages_test')
        ); 
```

ogizanagi: Here is what's used in Symfony's tests: 
https://github.com/symfony/symfony/blob/b285584e77e69ad9740711a12581c3fce46535ee/src/Symfony/Component/Translation/Tests/Extractor/PhpExtractorTest.php#L116

WhiteRabbitDE: Adjusted

ogizanagi:
Thank you! I really appreciate your help on this topic, as I don't have a windows machine to test easily.

PR #115, opened 2020-11-26
Run & fix inspections

ogizanagi:


composer.json
```diff
@@ -34,7 +34,8 @@
     "prefer-stable": true,
     "minimum-stability": "dev",
     "require": {
-        "php": ">=7.3"
+        "php": ">=7.3",
+        "symfony/symfony": "^4.4"
```
benji07, line 38: to remove ?

ogizanagi: indeed 👍

benji07 (approved):

PR #125, opened 2021-02-19
GraphQL bridge

mathroc:
@ogizanagi just made me realize I made [this merge request on my own fork](https://github.com/mathroc/PhpEnums/pull/1) a while ago ... :facepalm: 

so I’m moving it here.

Unfortunately I won’t have time to work on this at the moment, so feel free to use it any way you want if there’s something useful. otherwise I’ll look at what would be needed to merge this when my schedule allows it

### DIFF
--- a/src/Bridge/Overblog/GraphQL/EnumType.php
+++ b/src/Bridge/Overblog/GraphQL/EnumType.php
@@ -1,0 +1,89 @@
+<?php declare(strict_types=1);
+
+namespace Elao\Enum\Bridge\Overblog\GraphQL;
+
+use Elao\Enum\EnumInterface;
+use GraphQL\Type\Definition;
+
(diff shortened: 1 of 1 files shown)


ogizanagi:
For the ones interested:

> This is indeed something I'd like to add. Though I tried before without finding something satisfying enough.
Ideally, I'd like to generate the GraphQL types somehow, so we just have to declare the type in Overblog yaml files or whatever. (I'd just want to avoid creating a new class extending an abstract one for each enum class).
Is that something doable?

https://github.com/mathroc/PhpEnums/pull/1#issuecomment-782066335

ogizanagi:
Let's close. The feature is still interesting but need to be reworked for the new 2.x implementation with PHP 8.1 enum and might be supported in the future by `overblog/GraphQLBundle` directly.

Issue #44, opened 2018-02-20
document usage with easy admin bundle


tristanbes:
Hi 

a section on "'how to use with easy admin" could be nice when editing entities inside easy admin bundle.

ogizanagi:
Why not, PR welcomed :)

Actually it should be as easy as simply using any other form type (but as it's not a core form type, EasyAdminBundle does not provide an alias, so you have to use the FQCN).

See https://symfony.com/doc/master/bundles/EasyAdminBundle/book/edit-new-configuration.html#customize-the-form-fields

> - **type (optional):** the Symfony Form type used to render this field. You can use the short type name (e.g. email) instead of its fully qualified class name (e.g. Symfony\Component\Form\Extension\Core\Type\EmailType) even if your application runs on Symfony 3 (the needed conversion is done internally by the bundle). The allowed values are:
         - Any of the Symfony Form types.
         - Any of the custom EasyAdmin form types: easyadmin_autocomplete (they are explained later in this chapter).
> - **type_options (optional)**, a hash with the options passed to the Symfony Form type used to render the field.

About the EnumType provided in this package: https://github.com/Elao/PhpEnums#symfony-form-component

ogizanagi:
Sure. And you already opened #45 :)

But it should probably not be enough to close this issue. Showing a simple example with custom options and redirecting on the official EasyAdmin documentation still is relevant. Also, perhaps provide/showcase simple templates for list & show views rendering of enums and array of enums.

ogizanagi:
Closing as there is no activity here. If anyone is willing to enhance the documentation or open a PR to improve integration with EasyAdmin, please do.

PR #128, opened 2021-05-18
Document usage with easy admin 3 bundle

pluk77:
Issue #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Table of Contents
     * [Api-Platform](#api-platform)
       * [OpenApi / Swagger](#openapi--swagger)
     * [EasyAdmin](#easyadmin)
+      * [EasyAdmin2](#easyadmin2)
       * [EasyAdmin3](#easyadmin3) 
     * [JavaScript](#javascript)
     * [Twig](#twig)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -925,15 +926,34 @@ The library provides an `Elao\Enum\Bridge\ApiPlatform\Core\JsonSchema\Type\ElaoE
...
+  fields:
+    - { property: language, type: 'Elao\Enum\Bridge\Symfony\Form\Type\EnumType', type_options: { enum_class: 'App\Entity\Enum\LanguageEnum', required: true, }, label: 'Language', help: 'Select language from the drop down above'}
+```
+
+To ensure the listing is working, set the type to text.
+
+```yml
+list:
+  fields:
+    - { property: language, label: 'Language', type: 'text'}
```
ogizanagi, line 944: For this to work, the enum must be a readable enum. Also, it won't translate in case you use translation keys as readable values, nor null values I guess.

A solution for this is to create a custom template `templates/easyadmin/enum.html.twig`, like : 

```twig
{% if value is empty %}
    {{ include(entity_config.templates.label_null) }}
{% else %}
    {{ value|trans }}
{% endif %}
```

and use it like this:

```yml

```yml
list:
  fields:
    - { property: language, label: 'Language', template: easyadmin/enum.html.twig }
```

ogizanagi (commented):
Thank you for taking the time to contribute to this doc!
I never tried with EasyAdmin 3, so I trust you 👍🏻

README.md
pluk77: Thank you. I have included your comments in the patch. Hope that was your intention?

ogizanagi: That's fine, thank you :)

README.md
```diff
@@ -928,22 +928,43 @@ The library provides an `Elao\Enum\Bridge\ApiPlatform\Core\JsonSchema\Type\ElaoE
...
+{% if value is empty %}
+    {{ include(entity_config.templates.label_null) }}
+{% else %}
+    {{ value|trans }}
+{% endif %}
+```
+
+and use it like this:
+
+
```
ogizanagi, line 961:
```suggestion
```

ogizanagi:
Thanks @pluk77.

adpeyre:
Hi,

Is there a solution for an array of enums ?

"The value of the "modules" field of the entity with ID = "1" can't be converted into a string, so it cannot be represented by a TextField."

ogizanagi:
I don't know with EasyAdmin 3, but the EasyAdmin 2 solution for this is the same: use a custom template for rendering.
Perhaps something similar exist in EasyAdmin 3?

adpeyre:
No, I can't. EasyAdmin3 checks if value is a string. I think I've to create a custom field.

pluk77:
I indeed suspect a custom field is the only proper way to go. 

You could however create a new getter function in your entity that returns the array as an imploded sting. Not as nice as a custom field, but it works. https://symfony.com/doc/current/bundles/EasyAdminBundle/fields.html#unmapped-fields

If you do decide to create a custom field, will you share your code for others to re-use?

adpeyre:
Could be great a Symfony Bundle referencing extra types for EasyAdmin. :)

ogizanagi:
@adpeyre PR welcomed to add this in the enum bundle 😉

PR #131, opened 2021-05-27
Fix phpdoc return types

IonBazan:
This PR fixes type errors when performing PHPStan analysis on enums `static` should be used instead of `EnumInterface`.


ogizanagi (approved):

ogizanagi:
Thanks @IonBazan.

PR #132, opened 2021-05-27
Support default value for generated Doctrine types

### LINKED ISSUE #130: Default value for generated Doctrine types
Currently there is no way to set a default enum value when using the autogeneration of Doctrine types via configuration. The generated types extend `AbstractEnumType` but the type classes do not implement `onNullFromDatabase()` / `onNullFromPhp()`. Even if they did, they would not know how to retrieve a default value from the related enum class.

Thus I'd suggest an API to define a default value for an enum (e.g. using a new interface). The `AbstractEnumType` could then check for this and use it to implement the mentioned methods to provide defaults.

What do you think?

mbrodala:
Fixes #130


src/Bridge/Doctrine/DBAL/Types/TypesDumper.php
```diff
@@ -74,14 +80,47 @@ private function dump(array $types): string
...
                 break;
             case self::TYPE_STRING:
                 $baseClass = AbstractEnumType::class;
+
+                if ($defaultOnNull !== null) {
+                    $implementation .= <<<PHP
+
+            protected function onNullFromDatabase()
+            {
+                return \\{$defaultOnNull};
```
ogizanagi, line 115: As the configuration will contain the enum raw value, we should get an enum instance from it here using `EnumInterface::get`

ogizanagi (commented):
Looks good to me so far. Thank you for taking action here.

src/Bridge/Doctrine/DBAL/Types/TypesDumper.php
mbrodala: I'll give this a try but actually I tested this with an enum name like `App\Enum\MyEnum::FOO`.

I didn't think that people would want to put raw values into their configuration since that would mean having to change the configuration on enum value change.

ogizanagi: Yes of course, either using `App\Enum\MyEnum::FOO` or the value directly, it's the same. That's what I meant :)

mbrodala: OK, I've changed this now. I still needed to "unpack" the enum via `->getValue()` but that is to be expected.

src/Bridge/Doctrine/DBAL/Types/TypesDumper.php
```diff
@@ -101,21 +141,28 @@ private function getTypeCode(string $classname, string $enumClass, string $type,
...
+PHP;
+    }
 
-            public function getName(): string
-            {
-                return static::NAME;
+    private function valueAsCode($value): string
+    {
+        if (is_string($value)) {
+            if (strpos($value, '::') !== false) { // Constant name
```
mbrodala, line 154: I'm not totally happy with this heuristic since this is not bullet-proof. With this you cannot have `::` in enum values as `default`.

An alternative would be requiring a leading backslash in case of constant names but that would be less intuitive given that `types` is just fine without leading backslash. Also technically you could not have a leading backslash in enum values either anymore but that's even more of an edge case.

@ogizanagi What do you think?

mbrodala: Ah indeed, that applies here too. If documented this would actually be the best solution. :+1:

src/Bridge/Doctrine/DBAL/Types/TypesDumper.php
```diff
@@ -101,23 +141,26 @@ private function getTypeCode(string $classname, string $enumClass, string $type,
...
-                return static::NAME;
-            }
+$code
         }
     }
 
 PHP;
     }
 
+    private function valueAsCode($value): string
```
ogizanagi, line 151: I don't think there is a need for this method. Let's just keep the value as passed.

mbrodala: But then we'd get an PHP error (undefined constant) in case of a string value. We must at least add quotes to strings, otherwise they end up as toplevel PHP constants in the generated code.

mbrodala: I just noticed that I can just use `var_export()`. :facepalm:

ogizanagi: Ok, got it. I'm fine with the `var_export` for this purpose 👍🏻 

>  int, float and bool are valid enum value types AFAIK

Only int and string are officially valid enum types actually: https://github.com/Elao/PhpEnums/blob/0fbdccfe49ee47edecdfa91980e4e46b8584b8ff/src/EnumInterface.php#L18

mbrodala: Got it, thanks. I'll tackle the tests next.

PR #136, opened 2021-05-31
[CI] Add dedicated lint job + minor workflow tweaks

### LINKED ISSUE #135: [CI] Split the Code style checks into a dedicated build
So we have a dedicated check, separated from tests.

ogizanagi:
Fixes #135 with some more tweaks


.github/workflows/ci.yml
```diff
@@ -0,0 +1,123 @@
...
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: ${{ runner.os }}-composer-
+
+      - name: 'Install dependencies'
+        run: composer update --prefer-dist
+
+      - name: 'Check style'
+        run: vendor/bin/php-cs-fixer fix --dry-run --no-interaction --diff
```
IonBazan, line 40: I think it's better to put it in separate `yml` file and use https://github.com/marketplace/actions/oskar-php-cs-fixer to make the build faster.

ogizanagi: Thanks for the suggestion, but I prefer keeping it in sync with the dev workflow rather than using this image (not a strong argument, just a personal preference).
Keeping it in the same workflow is fine to me since it's still run in // and we don't need different triggers / scopes are relatively the same.

IonBazan:
Seems that GitHub is marking the commit as failed ❌  even when build has `continue-on-error: true` Maybe we should add conditional `${{ matrix.allow-fail && '|| true' }}` to the PHPUnit step 🤔

ogizanagi:
> Seems that GitHub is marking the commit as failed ❌ even when build has `continue-on-error: true` Maybe we should add conditional `${{ matrix.allow-fail && '|| true' }}` to the PHPUnit step 🤔

True, but we still require the check to appear as failed, otherwise we'll never spot a new failure (if we fix it later).
At least the workflow run is properly marked as successful.

Ideally, Github should provide something to exclude some checks from the commit.
Or we could update the check conclusion to `neutral`, but it's not allowed neither.

Perhaps only keep the bleeding edges job for PRs, but not on pushes. So we still get a feedback on this job per code change.

IonBazan:
Roger that!
BTW, Symfony 5.3 has been released, so we can use `symfony/phpunit-bridge` at `^5.3` 🎉

Issue #134, opened 2021-05-31
[CI] Fix the tests build for Windows


ogizanagi:
See https://github.com/Elao/PhpEnums/pull/133#discussion_r641388398

IonBazan:
@ogizanagi fixed in #137

PR #137, opened 2021-05-31
fixing Windows tests

IonBazan:
Turned out, setting git LF settings fixed the issue as by default, Windows runner's git is converting LF to CRLF while checking out the code, resulting invalid data to be produced. This PR solves this issue #134


ogizanagi:
Awesome, thank you for digging in!

PR #139, opened 2021-06-22
138: Fix Api Platform swagger documentation for  nullable enum

rcoiffard:


tests/Unit/Bridge/ApiPlatform/Core/JsonSchema/Type/ElaoEnumTypeTest.php
```diff
@@ -52,6 +52,7 @@ public function typeProvider(): iterable
                     0 => 'unknown',
                     1 => 'male',
                     2 => 'female',
+                    3 => null,
```
ogizanagi, line 55: The `can be a nullable array of enum` test case need to be updated as well :)

ogizanagi:
Thank you @rcoiffard.

PR #144, opened 2021-07-15
improve return types phpdocs

IonBazan:
Some of the return types were incorrect, missing the template `T` or missing an import.

### DIFF
--- a/src/ReadableEnum.php
+++ b/src/ReadableEnum.php
@@ -12,6 +12,11 @@ namespace Elao\Enum;
 
 use Elao\Enum\Exception\InvalidValueException;
 
+/**
+ * @template T of int|string
+ * @extends Enum<T>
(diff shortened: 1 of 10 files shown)


src/ReadableEnum.php
```diff
@@ -12,6 +12,11 @@
 
 use Elao\Enum\Exception\InvalidValueException;
 
+/**
+ * @template T of int|string
+ * @extends Enum<T>
+ * @implements ReadableEnumInterface<T>
```
IonBazan, line 18: This stuff is useful when creating the enums so that `phpstan` and `psalm` will detect the return type of `getValue()`, `values()`, etc.

ogizanagi (approved):

ogizanagi:
Thanks @IonBazan.

PR #145, opened 2021-07-15
Add Doctrine ODM bridge

### LINKED ISSUE #143: Add Doctrine ODM types support
It would be great to see Doctrine ODM integration here.
I've used my own enum types implementation in few projects and would like to share it in a PR which will be provided later this week.

Not sure how to handle the automatic Types generation when using as Symfony Bundle.

IonBazan:
Fixes #143 

This change introduces Doctrine ODM bridge.

- [x] Introduce basic enum support
- [x] Support for collection of enums 
- [x] Tests
- [x] Support for null values from php and database 
- [x] Readme changes
- [x] Types Dumper
- [x] Bundle integration

From what I see, it is not possible to handle `null` values in ODM in a custom way - they will alway be persisted and fetched as `null`: https://github.com/doctrine/mongodb-odm/blob/85d2b9f9664517ca0ece48547be3e68f7a2a2e8f/lib/Doctrine/ODM/MongoDB/Persisters/PersistenceBuilder.php#L71-L77

### DIFF
--- a/src/Bridge/Doctrine/ODM/Types/AbstractEnumType.php
+++ b/src/Bridge/Doctrine/ODM/Types/AbstractEnumType.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the "elao/enum" package.
+ *
+ * Copyright (C) Elao
+ *
+ * @author Elao <contact@elao.com>
+ */
+
+namespace Elao\Enum\Bridge\Doctrine\ODM\Types;
+
+use Doctrine\ODM\MongoDB\Types\ClosureToPHP;
+use Doctrine\ODM\MongoDB\Types\Type;
+use Elao\Enum\EnumInterface;
+
+/**
+ * @template TEnum of EnumInterface
+ */
+abstract class AbstractEnumType extends Type
+{
+    use ClosureToPHP;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value)
+    {
+        if (null === $value) {
+            return null;
+        }
+
+        $class = static::getEnumClass();
+
+        if ($value instanceof $class) {
+            return $value->getValue();
+        }
+
+        return $value;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @phpstan-return TEnum
+     */
+    public function convertToPHPValue($value): ?EnumInterface
+    {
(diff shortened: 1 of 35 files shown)


src/Bridge/Doctrine/ODM/Types/AbstractEnumType.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        if (null === $value) {
+            return $this->onNullFromPhp();
+        }
+
+        $class = static::getEnumClass();
+
+        if ($value instanceof $class) {
+            return $value->getValue();
+        }
```
IonBazan, line 57: This check is necessary because you may pass both enum and its value: `$qb->field('foo')->equals(Gender::MALE)` or `$qb->field('foo')->equals(Gender::MALE())`.

IonBazan:
@ogizanagi can you take a look on this when you're free before I move to the next steps?

ogizanagi:
Looks great so far. Do you plan to add some integration tests so we can test the implementations against a real ODM, as done for https://github.com/Elao/PhpEnums/blob/master/tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php for instance ?

IonBazan:
I will try to 😄  The logic behind the types is quite simple but there are many edge cases, it deserves to be tested thoroughly.
I will need to spin off a Mongo container in the CI to test the operations on DB.

IonBazan:
@ogizanagi I think this is ready for review - I will add readme instructions for non-bundle usage later today and implement TypesDumper and Bundle integration in a separate PR.

IonBazan:
@ogizanagi I managed to add ODM TypeDumper and integrate it with Symfony Bundle. I tested it with one of my existing projects and confirm it is working as expected 🚀 
I would suggest to split the readme into multiple files in `doc` directory as it becomes really long.

ogizanagi:
Hi @IonBazan , that's awesome, thanks for your reactivity. I may not be able to review it in the next days though.

I agree about the README.md, and we may rework it later. I might give [Stenope](https://stenopephp.github.io/Stenope/) a chance to generate this doc on Github Pages btw (👀 poke @Tom32i).

benji07 (approved):

ogizanagi:
Thanks @IonBazan for working on this feature, this is much appreciated.

PR #148, opened 2021-08-12
Improve PHPStan and Psalm annotations

B-Galati:
Hello!

I just upgrade the lib and notice some false positive with PHPStan analysis due to missing annotations.

So here are my fixes. Let me know if it's correct or not :-)

### DIFF
--- a/src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
+++ b/src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
@@ -14,12 +14,16 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Elao\Enum\EnumInterface;
 
+/**
+ * @template T of EnumInterface
+ */
 abstract class AbstractEnumType extends Type
 {
     /**
      * The enum FQCN for which we should make the DBAL conversion.
      *
-     * @return class-string<EnumInterface>
+     * @return string&EnumInterface
+     * @psalm-return class-string<T>
      */
     abstract protected function getEnumClass(): string;
 
@@ -46,7 +50,8 @@ abstract class AbstractEnumType extends Type
     /**
      * {@inheritdoc}
      *
-     * @param EnumInterface $value
+     * @param EnumInterface|null $value
+     * @psalm-param T|null $value
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
@@ -66,7 +71,6 @@ abstract class AbstractEnumType extends Type
             return $this->onNullFromDatabase();
         }
 
-        /** @var string|EnumInterface $class */
         $class = $this->getEnumClass();
 
         return $class::get($this->cast($value));
--- a/src/Enum.php
+++ b/src/Enum.php
@@ -25,6 +25,7 @@ abstract class Enum implements EnumInterface
      * This means you'll always get the exact same instance for a same enum value.
      *
      * @var EnumInterface<int|string>[]
+     * @psalm-var EnumInterface<T>[]
      */
     private static $instances;
 
@@ -37,7 +38,8 @@ abstract class Enum implements EnumInterface
     /**
      * The constructor is private and cannot be overridden: use the static get method instead.
      *
-     * @param mixed $value The raw value of an enumeration
+     * @param int|string $value The raw value of an enumeration
+     * @psalm-param T $value
      */
     final private function __construct($value)
     {
--- a/src/SimpleChoiceEnum.php
+++ b/src/SimpleChoiceEnum.php
@@ -22,7 +22,9 @@ namespace Elao\Enum;
  */
 abstract class SimpleChoiceEnum extends ReadableEnum
 {
+    /* @use AutoDiscoveredValuesTrait<T> */
     use AutoDiscoveredValuesTrait;
+    /* @use ChoiceEnumTrait<T> */
     use ChoiceEnumTrait {
         ChoiceEnumTrait::values insteadof AutoDiscoveredValuesTrait;
     }
--- a/src/AutoDiscoveredValuesTrait.php
+++ b/src/AutoDiscoveredValuesTrait.php
@@ -13,6 +13,8 @@ namespace Elao\Enum;
 use Elao\Enum\Exception\LogicException;
 
 /**
+ * @template T of int|string
+ *
  * Auto-discover enumerated values from public constants.
  */
 trait AutoDiscoveredValuesTrait
@@ -24,12 +26,10 @@ trait AutoDiscoveredValuesTrait
     private static $guessedReadables = [];
 
     /**
-     * @template T of int|string
-     *
      * @see EnumInterface::values()
      *
      * @return int[]|string[]
-     * @psalm-return list<T>
+     * @psalm-return T[]
      */
     public static function values(): array
     {
--- a/src/Bridge/Doctrine/ODM/Types/AbstractCollectionEnumType.php
+++ b/src/Bridge/Doctrine/ODM/Types/AbstractCollectionEnumType.php
@@ -32,7 +32,7 @@ abstract class AbstractCollectionEnumType extends CollectionType
     /**
      * {@inheritdoc}
      *
-     * @phpstan-return array<TEnum>|null
+     * @psalm-return array<TEnum>|null
      */
     public function convertToPHPValue($value)
     {
--- a/src/Bridge/Doctrine/ODM/Types/AbstractEnumType.php
+++ b/src/Bridge/Doctrine/ODM/Types/AbstractEnumType.php
@@ -42,7 +42,7 @@ abstract class AbstractEnumType extends Type
     /**
      * {@inheritdoc}
      *
-     * @phpstan-return TEnum
+     * @psalm-return TEnum
      */
     public function convertToPHPValue($value): ?EnumInterface
     {
--- a/src/ChoiceEnumTrait.php
+++ b/src/ChoiceEnumTrait.php
@@ -13,6 +13,8 @@ namespace Elao\Enum;
 use Elao\Enum\Exception\LogicException;
 
 /**
+ * @template T of int|string
+ *
  * Discover readable enumerated values by returning the enumerated values as keys and their labels as values
  * in {@link \Elao\Enum\ChoiceEnumTrait::choices()}, replacing the need to provide both:
  * - {@link \Elao\Enum\ReadableEnumInterface::readables()}
@@ -25,10 +27,8 @@ trait ChoiceEnumTrait
     /**
      * @see EnumInterface::values()
      *
-     * @template T of int|string
-     *
      * @return int[]|string[]
-     * @psalm-return list<T>
+     * @psalm-return T[]
      */
     public static function values(): array
     {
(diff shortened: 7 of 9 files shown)


src/Enum.php
```diff
@@ -14,7 +14,7 @@
 use Elao\Enum\Exception\LogicException;
 
 /**
- * @template T of int|string
+ * @template T
```
ogizanagi, line 17: Why removing `of int|string` it from here?
Is it enough for phpstan/psalm to understand since it `@implements EnumInterface<T>` which defines `@template T of int|string`?

ogizanagi (commented):
Great! Looking good to me at first glance, but @IonBazan reported me some days ago he got some issues with phpstan strict rules as well after upgrade. I believe it might not be included in this patch, since it was related to the ODM integration, or is it?

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
```diff
@@ -14,12 +14,17 @@
 use Doctrine\DBAL\Types\Type;
 use Elao\Enum\EnumInterface;
 
+/**
+ * @template T
```
IonBazan, line 18: I think the `T` should be an `EnumInterface` so for `Gender` enum it would look like:

```php
/**
 * @extends AbstractEnumType<Gender>
 */
class GenderType extends AbstractEnumType
```

then the `getEnumClass()` should `@psalm-return class-string<T>`

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
```diff
@@ -14,12 +14,17 @@
...
+ */
 abstract class AbstractEnumType extends Type
 {
     /**
      * The enum FQCN for which we should make the DBAL conversion.
      *
-     * @return class-string<EnumInterface>
+     * @return string&EnumInterface
+     * @phpstan-return class-string<EnumInterface<T>>
+     * @psalm-return class-string<EnumInterface<T>>
```
IonBazan, line 27: I don't think we should duplicate these annotations here. AFAIK PHPStan parses `@psalm-*` annotations too so we can just stick to psalm ones.

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
```diff
@@ -47,6 +52,8 @@ protected function onNullFromPhp()
      * {@inheritdoc}
      *
      * @param EnumInterface $value
+     * @psalm-param EnumInterface<T> $value
```
IonBazan, line 55: The value can actually be `null` here too.

IonBazan (commented):
Thanks for this contribution! I encountered same problems after update but didn't have time to fix that.

I was thinking whether we could annotate the fixture enums we keep in test directories and run phpstan against them to verify the annotations are correct?

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
B-Galati: I did not know :+1: Thanks I will try.

src/Enum.php
B-Galati: From my understanding, yes, it's enough to only keep it in the parent class.

I am not sure about psalm as I am not using it. So any feedback is welcome.

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
B-Galati: Do you have any reference ? It looks to work indeed :+1:

src/Enum.php
B-Galati: I just tried, it does not work as expected :D Si I reverted them all.

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
B-Galati: Indeed! Thanks!

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
B-Galati: Good catch, thanks!

B-Galati:
@ogizanagi @IonBazan It's ready for review again :-)

IonBazan (commented):
I think there are 2 missing spots:
https://github.com/Elao/PhpEnums/blob/c533ec1463fd8aa42b7afd771a9c7cd7070141b4/src/ReadableEnumInterface.php#L18 should be `@extends EnumInterface<T>`

otherwise it causes an error: 
> Class Elao\Enum\ReadableEnum specifies template type T of interface Elao\Enum\EnumInterface as int|string but it's already specified as string.

and https://github.com/Elao/PhpEnums/blob/c533ec1463fd8aa42b7afd771a9c7cd7070141b4/src/SimpleChoiceEnum.php#L21 should add:

```php
@uses AutoDiscoveredValuesTrait<T>
@uses ChoiceEnumTrait<T>
```

B-Galati:
@IonBazan Done! Thanks!

src/SimpleChoiceEnum.php
```diff
@@ -22,7 +22,9 @@
  */
 abstract class SimpleChoiceEnum extends ReadableEnum
 {
+    /* @use AutoDiscoveredValuesTrait<T> */
```
B-Galati, line 25: Perhaps I should have use `/**` instead? or `//` ? Let me know.

IonBazan: These annotations should be on the class level, right where you put `@extends`. It should also be `@uses` (with a **s**).

ogizanagi: It should be `/**` for docblocks indeed, but php-cs-fixer will revert it to `/*` since docblocks aren't usually allowed here.
I didn't find any documentation on `@use`, but shouldn't it rather be in the class's phpdoc using these traits?

IonBazan:
I've tested your branch and I had some issues regarding the `ReadableEnum`:
```
Return type (array<int|string, string>) of method App\Domain\Company\ReviewStatus::readables() should be covariant with return type (array<string, string>) of method Elao\Enum\ReadableEnumInterface<string>::readables()
```

Not sure where is that type `array<int|string, string>` as my `ReviewStatus` seems to be annotated correctly:
```php
/**
 * @extends RenderableEnum<string>
 */
final class ReviewStatus extends RenderableEnum
```

src/SimpleChoiceEnum.php
B-Galati: It's undocumented. I only found this: https://github.com/phpstan/phpstan-src/commit/87669234363d89252583664ef71309d5d8d3c7c1#diff-a2f75fb97bd3f8a99a1c56718a7ac23a03b9e6533eb4c6d34c4730f6d8e3132fR108.

For the phpdoc emplacement, for me I find it easier to set the `@use` juste above the place it's used to ease the maintenance: removing is no brainer as well as adding a new trait.

Moreover that's what is done in the commit I linked above.

B-Galati: And `@uses` does not work, at least with PHPStan.

B-Galati: Just opened an issue to document it: https://github.com/phpstan/phpstan/issues/5464

B-Galati:
> ```
> RenderableEnum
> ```

There is a typo, it's `ReadableEnum` and not `RenderableEnum`.

I tested it with this

```
/**
 * @extends ReadableEnum<string>
 */
final class ReviewStatus extends ReadableEnum
{

    public static function values(): array
    {
        return ['string'];
    }

    public static function readables(): array
    {
        return ['key' => 'value'];
    }
}
```

And it looks fine to me.

src/SimpleChoiceEnum.php
IonBazan: Thanks, that explains a lot!

IonBazan:
@B-Galati sorry, I forgot to mention extending a class adding extra features there but it's still the same issue just a bit more nested. If it's not reproducible by you, it might be just something wrong with my annotations. Thanks anyway!
```php
/**
 * @template T of int|string
 * @extends SimpleChoiceEnum<T>
 */
abstract class RenderableEnum extends SimpleChoiceEnum
```

IonBazan (approved):

B-Galati:
@IonBazan 

I just tried with this and it's ok:

```
/**
 * @template T of int|string
 * @extends SimpleChoiceEnum<T>
 */
abstract class RenderableEnum extends SimpleChoiceEnum
{

}

/**
 * @extends RenderableEnum<string>
 */
final class Test extends RenderableEnum
{

}
```

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
```diff
@@ -14,12 +14,16 @@
...
+/**
+ * @template T of EnumInterface
+ */
 abstract class AbstractEnumType extends Type
 {
     /**
      * The enum FQCN for which we should make the DBAL conversion.
      *
-     * @return class-string<EnumInterface>
+     * @return string&EnumInterface
```
benji07, line 25: this line seems wrong to me

ogizanagi: It's wrong indeed, but it's on purpose for autocompletion to work here: https://github.com/Elao/PhpEnums/blob/c533ec1463fd8aa42b7afd771a9c7cd7070141b4/src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php#L72

but not sure if we should keep it here on the method signature or locally on usage (or remove completly)

ogizanagi:
Let's go with this. Thank you very much to both of you @B-Galati and @IonBazan 🙏🏻

B-Galati:
Thank you all for your reviews :+1: And thanks for the quick release :+1:

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
bendavies: is `AbstractEnumSQLDeclarationType` missing an annotation?

bendavies: for example, how should i annotate this class?

```php
final class FormUsageEnumType extends AbstractEnumSQLDeclarationType
{
    public const NAME = 'form_usage';

    protected function getEnumClass(): string
    {
        return FormUsage::class;
    }

    public function getName(): string
    {
        return self::NAME;
    }
}
```

phpstan will complain:

```Method FormUsageEnumType::getEnumClass() return type with generic interface Elao\Enum\EnumInterface does not specify its types: T
```

bendavies: should `AbstractEnumSQLDeclarationType` have 
```php

/**
 * @template T of EnumInterface
 * @extends AbstractEnumType<T>
 */
abstract class AbstractEnumSQLDeclarationType extends AbstractEnumType {}
```

?

B-Galati: I am doing something like this 


```
/**
 * @extends AbstractEnumType<Status>
 */
final class StatusType extends AbstractEnumType
{
    private const NAME = 'status';

    /** {@inheritdoc} */
    protected function getEnumClass(): string
    {
        return Status::class;
    }

    /** {@inheritdoc} */
    public function getName(): string
    {
        return static::NAME;
    }

    /** {@inheritdoc} */
    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform): string
    {
        return 'TEXT';
    }
}
```

PR #150, opened 2021-08-20
fix collection enum types for ODM

IonBazan:
I didn't test properly the type for enum collections in ODM previously and just found an issue where the collection is not converted to enums during hydration due to missing `ClosureToPHP` trait. It tells the ODM how to generate the hydrator and default behaviour omits the `convertToPHPValue` call.

The singular enum type was working as it already uses that trait.


ogizanagi:
Thanks for fixing this bug @IonBazan.

PR #152, opened 2021-08-20
handle readable enums which contain values which need escaping in javascript

bendavies:


composer.json
```diff
@@ -40,6 +40,7 @@
     "prefer-stable": true,
     "require": {
         "php": ">=7.3",
+        "ext-json": "*",
```
bendavies, line 43: is this a BC break?

PR #153, opened 2021-08-24
add native jsonSerialize

IonBazan:
It should be possible to use good old `json_encode` to serialize the enum.

```php
echo json_encode(\Elao\Enum\Tests\Fixtures\Enum\Gender::MALE());
```

Before: `{}`
After: `"male"`

### DIFF
--- a/src/Enum.php
+++ b/src/Enum.php
@@ -12,12 +12,13 @@ namespace Elao\Enum;
 
 use Elao\Enum\Exception\InvalidValueException;
 use Elao\Enum\Exception\LogicException;
+use JsonSerializable;
 
 /**
  * @template T of int|string
  * @implements EnumInterface<T>
  */
-abstract class Enum implements EnumInterface
+abstract class Enum implements EnumInterface, JsonSerializable
 {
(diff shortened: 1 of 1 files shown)


ogizanagi:
Hi @IonBazan . I'm not sure I'd like to provide this feature out-of-the-box. The format would be arbitrary and one might also render differently readable enums for instance. Of course this could be overwritten, but then I don't see much value into providing this default implementation here? WDYT?
This could also lead to unexpected behaviors / minor BC-breaks in some place were JsonSerializable objects are handled, without the user noticing it.

IonBazan:
You are right, the `JsonSerializable` introduces a lot of magic but on the other hand, most of the time you expect it to serialize into string instead of empty object. I ended up adding `jsonSerialize` to my enums just in case they will be serialized somewhere which should be provided by this lib.

Moreover, `myclabs/php-enum` implements that interface too: https://github.com/myclabs/php-enum/blob/db0158242788c972cd137780a278d292eaaa6e6c/src/Enum.php#L314

PR #163, opened 2021-11-15
[CI] Prepare for next SF & PHP versions

ogizanagi:
Symfony 5.4/6.0 & PHP 8.1 should not be allowed to fail in two weeks.

---

I don't know yet why the 8.1 builds are failing with:

> Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "`@ORM\Entity`" in class App\Entity\User was never imported. Did you maybe forget to add a "use" statement for this annotation?

Everything seems fine locally with PHP 8.1 (`PHP 8.1.0-dev (cli) (built: Nov 15 2021 00:35:22) (NTS)`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,50 +57,58 @@ jobs:
             mongodb: true
 
           # Previous Symfony versions
-          - name: 'Test 5.2 Symfony [Linux, PHP 8.0]'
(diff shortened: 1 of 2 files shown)


ogizanagi:
@chalasr Ok, thx. I knew you'll get me the answer 😇

ogizanagi:
All green 💚 , as planned 🤘🏻

PR #170, opened 2022-01-13
Allow to provide extra values to EnumCase attribute

ogizanagi:
cc @IonBazan 

https://github.com/Elao/PhpEnums/blob/enum-case-extras/README.md#extra-values

Would this fit to your needs?

### DIFF
--- a/src/EnumCaseAttributesTrait.php
+++ b/src/EnumCaseAttributesTrait.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the "elao/enum" package.
+ *
+ * Copyright (C) Elao
+ *
+ * @author Elao <contact@elao.com>
+ */
+
+namespace Elao\Enum;
+
+use Elao\Enum\Attribute\EnumCase;
+
+trait EnumCaseAttributesTrait
+{
+    private function getEnumCaseAttribute(): ?EnumCase
+    {
+        return static::enumCaseAttributes()[$this] ?? null;
+    }
+
+    /**
+     * @return \SplObjectStorage<\UnitEnum,EnumCase>
+     */
+    private static function enumCaseAttributes(): \SplObjectStorage
+    {
+        static $attributes;
+
+        if (!isset($attributes)) {
+            $attributes = new \SplObjectStorage();
+
+            foreach ((new \ReflectionEnum(static::class))->getCases() as $rCase) {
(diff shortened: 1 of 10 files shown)


IonBazan:
Makes sense to me 💯 
Would that be also possible to add a method to get the actual attribute instance to allow extending `EnumCase` attribute while still benefiting from the cache?

ogizanagi:
> Would that be also possible to add a method to get the actual attribute instance to allow extending EnumCase attribute while still benefiting from the cache?

Might be. But I think your use-case might be covered by:

```php
/**
 * @final
 */
#[\Attribute(\Attribute::TARGET_CLASS_CONSTANT)]
class RenderableCase extends EnumCase
{
    public function __construct(string $label, string $color, string $icon)
    {
        parent::__construct($label, extra: compact('color', 'icon'))
    }
}
```

while still requiring only `getExtra`, if that's only about more scoped PHP attributes, isn't it?

IonBazan:
My concern is type safety. With this `extra: ['something' => 'test']` we are not sure what items and what types are stored there. If we could have a method that returns all attributes that extend `EnumCase`, we could store and retrieve any kind of extra information with ease and type safety.

Something like:
```php
/**
 * @return EnumCase[]
 */
public function getAttributes(): array
{
    // ...
}

public function getColor(): ?string
{
    foreach ($this->getAttributes() as $attribute) {
        if ($attribute instanceof HasColor) {
            return $attribute->color;
        }
    }

    return null;
}
```

We could combine both approaches - for quick results, developers might use this `extra` field, but for more sophisticated logic, they might want to use the helper that returns the attributes assigned to the case.

ogizanagi:
But the example in https://github.com/Elao/PhpEnums/pull/170#issuecomment-1012056988 is enough to provide type safety to me (unless you use both `EnumCase` and `RenderableCase` in your code base for the same needs, but that's not the point of creating a dedicated attribute then).

src/EnumCaseAttributesTrait.php
```diff
@@ -0,0 +1,48 @@
...
+ * @author Elao <contact@elao.com>
+ */
+
+namespace Elao\Enum;
+
+use Elao\Enum\Attribute\EnumCase;
+
+trait EnumCaseAttributesTrait
+{
+    private function getEnumCaseAttribute(): ?EnumCase
```
ogizanagi, line 19: (methods are private, but can be used by any userland implementation using this trait, as not marked `@internal`)

benji07 (approved):

PR #177, opened 2022-01-28
add Faker provider

IonBazan:
This PR adds simple Faker integration. Since `nelmio/alice` allows to use `<(My\Awesome\Suit::HEART)>` notation, there is no point for `enum()` generator but I think `randomEnum()` is still valid.

Registering the provider in v1 required some manual work in `services.yaml`. Do you think we could automate that with bundle configuration like this:

```yaml
elao_enum:
  nelmio_alice:
    aliases:
       Suit: App\Enums\Suit
       Permissions: App\Enums\Permissions
```

Would that make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,3 +437,60 @@ class Card
     private Suit $field;
 }
 ```
+
+### Faker
+
+The PhpEnums library provides a faker `EnumProvider` allowing to select random enum cases:
+
+```php
+use \Elao\Enum\Bridge\Faker\Provider\EnumProvider;
+ 
+$faker = new Faker\Generator();
+$faker->addProvider(new EnumProvider());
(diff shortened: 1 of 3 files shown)


ogizanagi:
I think there is not much benefit to have a bundle config for this, since registering the provider in Alice is done this way:

```yaml
services:
  Elao\Enum\Bridge\Faker\Provider\EnumProvider:
    arguments:
      - Suit: App\Enums\Suit
        Permissions: App\Enums\Permissions
  tags: ['nelmio_alice.faker.provider']
```

which is simple enough not to bother about introducing another conditional config entry.

ogizanagi:
Thanks @IonBazan.

PR #176, opened 2022-01-28
Add Doctrine ODM integration

IonBazan:
Adding Doctrine ODM integration:

- [x] Add simple base type
- [x] Add collection base type
- [x] Add tests for base types
- [x] Add types dumper
- [x] Integrate with bundle
- [x] Add tests for the dumper
- [x] Add tests for the bundle integration

Most tests could be copied from the v1 and adjusted slightly.

Please note that I added `type` option for each Doctrine Type - for now, ORM will only support `single` as `collection` is not implemented yet.

Issue #124, opened 2021-02-18
Prepare for PHP 8.1 Enum RFC changes


ogizanagi:
As per https://wiki.php.net/rfc/enumerations#backward_incompatible_changes:

> “enum” becomes a language keyword, with the usual potential for naming conflicts with existing global constants and class/interface/trait names.

So we'll at least need to rename https://github.com/Elao/PhpEnums/blob/master/src/Enum.php

This new enum type is great news. Note this will not invalidate this library though, since we provide extra integrations that can still be helpful, at least for some times. Frameworks and other packages are likely to provide their own. We'll see if a rewrite relying on this new type should land or not.

ogizanagi:
https://github.com/symfony/symfony/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+enum+label%3A%22Help+wanted%22

Likely to be dropped in the future, if any rewrite of this library is done based on the new enum type:

- [Serializer integration](https://github.com/Elao/PhpEnums#symfony-serializer-component) => https://github.com/symfony/symfony/issues/40241
- DBAL Types & DBAL types generator
- HTTP Kernel [Controller Argument Resolver](https://github.com/Elao/PhpEnums#symfony-httpkernel-component)
- [API Platform](https://github.com/Elao/PhpEnums#api-platform): no specificity for readable or flagged enum? Anything else to consider?
- [Faker](https://github.com/Elao/PhpEnums#faker): could be contributed to [FakerPHP](https://github.com/FakerPHP) directly for native enum support
- [Validator constraint](https://github.com/Elao/PhpEnums#symfony-validator-component). Likely to be supported in Symfony directly. => https://github.com/symfony/symfony/pull/43047


Might still benefits:
- [VarDumper caster](https://github.com/Elao/PhpEnums#symfony-vardumper-component) for the readable & flagged enums (Symfony will support basic enums => https://github.com/symfony/symfony/issues/40238)
- [Symfony Form Type](https://github.com/Elao/PhpEnums#symfony-form-component) ? For readable & flagged enum specificities? => https://github.com/symfony/symfony/pull/43095
- [Javascript generator](https://github.com/Elao/PhpEnums#javascript)
- [Twig extension](https://github.com/Elao/PhpEnums#twig) ? At least for readables?
- [translation extractor & config](https://github.com/Elao/PhpEnums#readable-enums) for readables

Extras:
[...]

PR #179, opened 2022-04-09
Add SQL Enum support

### LINKED ISSUE #178: Use actual MySQL ENUM type with Doctrine DBAL
Hi, thanks so much for your package. I do not have time to actually make a PR at the moment, but I wanted to share what I did to support native MySQL ENUM types.

I have my own AbstractEnumType class:

```php
<?php

namespace App\DBAL;

use BackedEnum;
use Doctrine\DBAL\Platforms\AbstractPlatform;
use Doctrine\DBAL\Types\Type;

abstract class AbstractEnumType extends Type
{
    abstract public static function getEnumsClass(): string;

    public function getName(): string // the name of the type.
    {
        return static::getEnumsClass();
    }

    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform): mixed
    {
        /** @var BackedEnum $class */
        $class = static::getEnumsClass();

        $values = array_map(
        /** @var BackedEnum $val */
            function ($val) {
                return "'" . $val->value . "'";
            },
            $class::cases()
        );

        return "ENUM(" . implode(", ", $values) . ")";
    }

    public function convertToDatabaseValue($value, AbstractPlatform $platform): mixed
    {
        if ($value instanceof \BackedEnum) {
            return $value->value;
        }
        return null;
    }

    public function convertToPHPValue($value, AbstractPlatform $platform): mixed
    {
        if (false === enum_exists($this->getEnumsClass(), true)) {
            throw new \LogicException("This class should be an enum");
        }

        /** @var BackedEnum $class */
        $class = static::getEnumsClass();

        return $class::tryFrom($value);
    }


    /**
     * @codeCoverageIgnore
     */
    public function requiresSQLCommentHint(AbstractPlatform $platform): bool
    {
        return true;
    }
}
```

When I make new enum like

```php
<?php

namespace App\Entity\Enum;

use Elao\Enum\Attribute\EnumCase;
use Elao\Enum\ReadableEnumInterface;
use Elao\Enum\ReadableEnumTrait;

enum AdminRole: string implements ReadableEnumInterface
{
    use ReadableEnumTrait;

    #[EnumCase('Admin')]
    case ADMIN = 'ROLE_ADMIN';

    #[EnumCase('Super Admin')]
    case SUPER_ADMIN = 'ROLE_SUPER_ADMIN';

    #[EnumCase('CSR')]
    case CSR = 'ROLE_CSR';

    #[EnumCase('Manager')]
    case MANAGER = 'ROLE_MANAGER';

    #[EnumCase('Accounting')]
    case ACCOUNTING = 'ROLE_ACCOUNTING';

    #[EnumCase('SEO')]
    case SEO = 'ROLE_SEO';
}
```

I then make also a "wrapper" in DBAL folder:

```php
<?php

namespace App\DBAL;

use App\Entity\Enum\AdminRole;

class AdminRoleType extends AbstractEnumType
{
    public static function getEnumsClass(): string // the enums class to convert
    {
        return AdminRole::class;
    }
}
```

This wrapper just returns the PHP enum which it links.

In `config/services.yaml` I put:

```
services:
   _instanceof:
        App\DBAL\AbstractEnumType:
          tags: ['app.doctrine_enum_type']
```

and then in Kernel.php, I register all "wrapper" classes as valid types:

```php
<?php

namespace App;

use App\DBAL\AbstractEnumType;
use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\HttpKernel\Kernel as BaseKernel;

class Kernel extends BaseKernel implements CompilerPassInterface
{
    use MicroKernelTrait;

    public function process(ContainerBuilder $container): void
    {
        $typesDefinition = [];
        if ($container->hasParameter('doctrine.dbal.connection_factory.types')) {
            $typesDefinition = $container->getParameter('doctrine.dbal.connection_factory.types');
        }
        $taggedEnums = $container->findTaggedServiceIds('app.doctrine_enum_type');

        foreach ($taggedEnums as $enumType => $definition) {
            /** @var $enumType AbstractEnumType */
            $typesDefinition[$enumType::getEnumsClass()] = ['class' => $enumType];
        }

        $container->setParameter('doctrine.dbal.connection_factory.types', $typesDefinition);
    }

}
```

And thats it!. In Entity I use:

```php
    #[ORM\Column(type: AdminRole::class)]
    private AdminRole $admin_role = AdminRole::CSR;
```

and the SQL generated by Doctrine migrations is:

```sql
CREATE TABLE admin (id INT AUTO_INCREMENT NOT NULL, email VARCHAR(180) NOT NULL, admin_role ENUM('ROLE_ADMIN', 'ROLE_SUPER_ADMIN', 'ROLE_CSR', 'ROLE_MANAGER', 'ROLE_ACCOUNTING', 'ROLE_SEO') NOT NULL COMMENT '(DC2Type:App\\Entity\\Enum\\AdminRole)', password VARCHAR(255) NOT NULL, token VARCHAR(255) DEFAULT NULL, UNIQUE INDEX UNIQ_880E0D76E7927C74 (email), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB
```

Adding new native ENUM type consists of only creating the php enum, then making the wrapper class.

So I was thinking this might be interesting to extend in your own AbstractEnumType doctrine bundle, which now uses either `INT` or `VARCHAR`. And if nothing else, it might help others who want native MySQL ENUMs

Thanks for your work!

michnovka:
Backport the feature originally present in 1.x for native ENUM


michnovka:
I have made some changes to .gitignore, to ignore .idea files (I assume I cant be the only one working in PHPStorm) as its standard used by both Symfony and Doctrine.

I have also included phpunit and psalm dependencies as dev, because well we have lot of @psalm annotations, and then because we actually do have tests for this bundle, so it makes no sense not to have phpunit included.

Speaking of tests, I would appreciate some help on writing them, as I am still a bit lost as to how this bundle is structured. We can start by writing down list of tests that need to be implemented and move on from there:

1. Create schema with ENUM type
2. Check retrieval and cast of values from SQL DB ENUM type into PHP representation
3. Check saving of values from PHP into SQL DB
4. Check configuration option `doctrine.type.enum_sql_declaration` works as expected (enabled ENUM SQLs by default)
5. Check configuration option doctrine.type.types.type = enum works as expected

Possible considerations/discussion:
1. I have made the option to use INT ENUMs when its BackedEnum of int type. I.e. to have ENUM(1,2,3) instead of ENUM('1','2','3'). Yet if its FlagBag, I keep the INT type in DB. Is this desired? For that reason I had to change `Elao\Enum\Bridge\Doctrine\DBAL\Types\AbstractEnumType::isIntBackedEnum` from private to protected.
2. Do we want ODM support too? I dont use ODM in my life, so I have literally 0 experience with it.
3. Are we happy with the type name `SINGLE` ? Maybe using `PRIMITIVE` or `SIMPLE` has more sense?
4. Before we make the final release of 2.x maybe it would be interesting to consider removing all `const` values used inside the bundle into their own Enums? After all this is an enum package :) e.g. `Elao\Enum\Bridge\Doctrine\DBAL\Types\TypesDumper::TYPE_SINGLE` could be made into an Enum type. Checked the code for `const`s and `TypeDumper` seems to be the only place, so I could make that change as part of this PR

ogizanagi:
> I have made some changes to .gitignore, to ignore .idea files (I assume I cant be the only one working in PHPStorm) as its standard used by both Symfony and Doctrine.

You're not the only one 😉 But the good practice actually is to register these specific files in a global `.gitignore` file in your user's home directory, not the project one. See https://gist.github.com/subfuzion/db7f57fff2fb6998a16c for instance.

> I have also included phpunit and psalm dependencies as dev, because well we have lot of https://github.com/psalm annotations, and then because we actually do have tests for this bundle, so it makes no sense not to have phpunit included.

Please revert these changes, since these are not necessary. We already have PHPUnit installed through the Symfony phpunit-bridge. Use `vendor/bin/simple-phpunit` to execute the tests.
If you have `make` and Symfony CLI installed on your machine, you can run `make test`. See the [Makefile](https://github.com/Elao/PhpEnums/blob/2.x/Makefile)

About Psalm, the annotations were added for using psalm/phpstan on userland code, we do not have any CI task enforcing it yet ; it should be added in another PR along with proper config & CI.

---

About tests, some existing resources:


>1. Create schema with ENUM type

The [integration tests](https://github.com/Elao/PhpEnums/tree/2.x/tests/Integration) are using a Symfony app to configure some enum using the bundle and the ORM config:

https://github.com/Elao/PhpEnums/blob/2.x/tests/Fixtures/Integration/Symfony/config/config.yml#L19-L33

Then, an integration test case class using the Symfony Kernel make some assertions by consuming the app:

https://github.com/Elao/PhpEnums/blob/2.x/tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php

There are also unit tests for this and the following points for:
- [the DBAL type itself](https://github.com/Elao/PhpEnums/blob/2.x/tests/Unit/Bridge/Doctrine/DBAL/Types/EnumTypeTest.php)
- the [TypesDumper generated code](https://github.com/Elao/PhpEnums/blob/2.x/tests/Unit/Bridge/Doctrine/DBAL/Types/TypesDumperTest.php)

>2. Check retrieval and cast of values from SQL DB ENUM type into PHP representation

See [this integration test](https://github.com/Elao/PhpEnums/blob/49a73e0d11684318e11ad968731a83b79f55fead/tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php#L44-L54) and [unit test](https://github.com/Elao/PhpEnums/blob/2.x/tests/Unit/Bridge/Doctrine/DBAL/Types/EnumTypeTest.php)

>3. Check saving of values from PHP into SQL DB

See [this integration test](https://github.com/Elao/PhpEnums/blob/49a73e0d11684318e11ad968731a83b79f55fead/tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php#L56-L69) and [unit test](https://github.com/Elao/PhpEnums/blob/2.x/tests/Unit/Bridge/Doctrine/DBAL/Types/EnumTypeTest.php)

>4. Check configuration option `doctrine.type.enum_sql_declaration` works as expected (enabled ENUM SQLs by default)

Such tests are made with unit tests:

- [testing the Configuration class](https://github.com/Elao/PhpEnums/blob/49a73e0d11684318e11ad968731a83b79f55fead/tests/Unit/Bridge/Symfony/Bundle/DependencyInjection/ConfigurationTest.php#L36-L58)
- [testing the bundle Extension](https://github.com/Elao/PhpEnums/blob/49a73e0d11684318e11ad968731a83b79f55fead/tests/Unit/Bridge/Symfony/Bundle/DependencyInjection/ElaoEnumExtensionTest.php#L35-L64)

>5. Check configuration option doctrine.type.types.type = enum works as expected

this should be tested in both unit tests case mentioned below (config & extension).

---

> 1. I have made the option to use INT ENUMs when its BackedEnum of int type. I.e. to have ENUM(1,2,3) instead of ENUM('1','2','3'). Yet if its FlagBag, I keep the INT type in DB. Is this desired? For that reason I had to change `Elao\Enum\Bridge\Doctrine\DBAL\Types\AbstractEnumType::isIntBackedEnum` from private to protected.

I think sql enum should preferably be stored as string. So let's always use `ENUM('1','2','3')`.

About `FlagBag`, you should not worry about it in this PR, since it would be a dedicated DBAL type storing the combination of multiple enum cases into 1 INT column. So you cannot store it using an `ENUM` SQL type.

> 2. Do we want ODM support too? I dont use ODM in my life, so I have literally 0 experience with it.

ODM is already supported since #176. There is no change to made in tis PR regarding a "native" enum type in MongoDB.

> 3. Are we happy with the type name `SINGLE` ? Maybe using `PRIMITIVE` or `SIMPLE` has more sense?

I'd go with `SCALAR`

4. Before we make the final release of 2.x maybe it would be interesting to consider removing all `const` values used inside the bundle into their own Enums? After all this is an enum package :) e.g. `Elao\Enum\Bridge\Doctrine\DBAL\Types\TypesDumper::TYPE_SINGLE` could be made into an Enum type. Checked the code for `const`s and `TypeDumper` seems to be the only place, so I could make that change as part of this PR

That's a bit out-of-scope, so I'd say please avoid doing this in this PR 😄 Anyway, keep in mind not every const values make sense as enums. Regarding the TypesDumper ones, it doesn't bring much value, mostly because it's internal as well as using enums instead might complexify a bit the Symfony DI config handling which does not support backed enums out-of-the-box.

michnovka:
Good to know about that global .gitignore, I learnt something new.

I have removed the newly added dependencies, except for doctrine-odm-bundle. This is vital for phpunit tests of ODM. If I dont add it, I always get errors when running tests. Psalm and phpunit gone.                                  

I have removed int ENUMs, so they are always strings.

Did not yet rename SINGLE to SCALAR, I will do so once we agree on it globally. Or does your ring rule them all @ogizanagi ? :)

Will have a look at the remaining points and mainly tests later.

ogizanagi:
>  have removed the newly added dependencies, except for doctrine-odm-bundle. This is vital for phpunit tests of ODM

That's unexpected : the test cases should be skipped when the bundle is not installed 🤔 :

https://github.com/Elao/PhpEnums/blob/49a73e0d11684318e11ad968731a83b79f55fead/tests/Integration/Bridge/Doctrine/ODM/Type/EnumTypeTest.php#L28-L30

> Did not yet rename SINGLE to SCALAR, I will do so once we agree on it globally. 

Let's go with `SCALAR` for DBAL types yes (since we'll also have `enum` and later on: `flagbag`, `json_collection`, `csv_collection`). For ODM ones, we'll keep `SINGLE` one as opposed to `COLLECTION`, as already implemented and since there should not be any other value.

michnovka:
> That's unexpected : the test cases should be skipped when the bundle is not installed thinking

Ah, I didnt realize its test skip not an error. So this is desired? I think that if we have tests for that part of bridge, then why not require it as dev? Anyways removing since I am not touching ODM so it should not be part of this PR

ogizanagi:
> > That's unexpected : the test cases should be skipped when the bundle is not installed thinking
> 
> Ah, I didnt realize its test skip not an error. So this is desired? I think that if we have tests for that part of bridge, then why not require it as dev? Anyways removing since I am not touching ODM so it should not be part of this PR

The reasoning here simply is testing ODM requires an advanced setup on your machine (php mongodb extension, a mongodb instance running, …), so we won't impose it be default and simply skip the related test cases when the ODM bundle is not installed (it's installed on CI in the github actions, so any PR is fully tested).

michnovka:
Working on the tests and I have some confusion in https://github.com/Elao/PhpEnums/blob/49a73e0d11684318e11ad968731a83b79f55fead/tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php#L56

Why does saving a `Card` entity with `Card.suit = null` result in `Spades` in DB and on retrieval? The `Card.suit` field is nullable, null is therefore an accepted value

michnovka:
I came across one "issue" or better said property of current `TypesDumper`.

I tried this config:
```yaml
elao_enum:
    doctrine:
        types:
            App\Enum\Suit:
                default: !php/const App\Enum\Suit::Spades
            suit_sql_enum:
                class: App\Enum\Suit
                type: enum
```

So I wanted to have an alias for the same Enum. This resulted in this code by `TypesDumper`

```php
<?php

namespace ELAO_ENUM_DT_DBAL\App\Enum {

    if (!\class_exists(SuitType::class)) {
        class SuitType extends \Elao\Enum\Bridge\Doctrine\DBAL\Types\AbstractEnumType
        {
            protected function getEnumClass(): string
            {
                return \App\Enum\Suit::class;
            }

            protected function onNullFromDatabase(): ?\BackedEnum
            {
                return \App\Enum\Suit::from('S');
            }

            protected function onNullFromPhp(): int|string|null
            {
                return 'S';
            }
        }
    }

    if (!\class_exists(SuitType::class)) {
        class SuitType extends \Elao\Enum\Bridge\Doctrine\DBAL\Types\AbstractEnumSQLDeclarationType
        {
            protected function getEnumClass(): string
            {
                return \App\Enum\Suit::class;
            }

            public function getName(): string
            {
                return 'suit_sql_enum';
            }
        }
    }

}
```

So the issue is clear. The name for the types is based on the Enum class name. I think it should be based instead on the type name as specified in config and slugged. Opinions?

ogizanagi:
You're right indeed, using the PascalCased type name might be better for such edge-cases.
Just duplicate the enum class for your test-case if you need it for now, in case you prefer not handling with this in this PR right now. We'll fix it in a dedicated one later.

michnovka:
I am fine with doing it in this one. Can we use https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/namingstrategy.html since Doctrine is needed anyways? Or do you prefer I make my own toPascalCase function?

michnovka:
@ogizanagi Running into issues with tests with SQLite... It has no ENUM type. What do you suggest? I have checked the original test in 1.x and there seems to be no testing of actual SQLEnum type functionality

michnovka:
There are further issues with this. Because the test requires use of SchemaManager to create the DB, I am getting the following errors when making the tests:

```
Doctrine\DBAL\Exception : Unknown database type enum requested, Doctrine\DBAL\Platforms\MariaDb1027Platform may not support it.
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Platforms/AbstractPlatform.php:396
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/MySQLSchemaManager.php:137
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:981
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:208
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:321
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/MySQLSchemaManager.php:336
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:306
 /home/superuser/scripts/PhpEnums/vendor/doctrine/dbal/src/Schema/AbstractSchemaManager.php:1229
 /home/superuser/scripts/PhpEnums/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/SchemaTool.php:846
 /home/superuser/scripts/PhpEnums/vendor/doctrine/orm/lib/Doctrine/ORM/Tools/SchemaTool.php:831
 /home/superuser/scripts/PhpEnums/tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php:33
 ```

I think this is the reason why we didnt have tests in 1.x. I am not sure what we can do fix this, any ideas?

**Comment by michnovka** [comment]
I have pushed the changes so you can have a look too. But I dont think I can resolve this. The original version also did not have tests and I believe the main issue is due to SQLite lack of support. However even if we use locally MySQL, I have the problems with SchemaManager trying to map enum type, which is not supported by the platform officially. I do not have this issue when using migrations / working with ORM only, so I assume it has to do with the SchemaManager itself. But please, a review is needed.

In the meantime I have made the SINGLE -> SCALAR change. I have also changed how the TypesDumper creates names - if no custom name is specified then nothing changes. But if custom name is used in config, then its pascal-cased and used instead of the base name of the enum

The tests I included in this PR are marked as skipped for the above mentioned issues

**Comment by michnovka** [comment]
@ogizanagi Can we merge this?

**Comment by michnovka** [comment]
@ogizanagi fixed lint errors, please re-run tests

File: src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php [hunk 1]
```diff
@@ -37,6 +37,10 @@ private function addDoctrineDbalSection(ArrayNodeDefinition $rootNode)
             ->addDefaultsIfNotSet()
             ->fixXmlConfig('type')
             ->children()
+                ->booleanNode('enum_sql_declaration')
+                    ->defaultValue(false)
+                    ->info('If true, by default for enumerations, generate DBAL types with an ENUM SQL declaration with enum values instead of a VARCHAR/INT (Your platform must support it)')
```
ogizanagi:
```suggestion
                    ->info('If true, generate DBAL types with an ENUM SQL declaration with enum values instead of a VARCHAR/INT (Your platform must support it)')
```

src/Bridge/Doctrine/DBAL/Types/AbstractEnumSQLDeclarationType.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform): string
+    {
+        if (!$platform instanceof AbstractMySQLPlatform) {
+            throw new LogicException('ENUMs not supported on current platform');
+        }
+
+        $values = array_map(
+            function ($val) {
+                return "'" . $val->value . "'";
```
ogizanagi:
```suggestion
                return "'{$val->value}'";
```

src/Bridge/Doctrine/DBAL/Types/AbstractEnumSQLDeclarationType.php
```diff
@@ -0,0 +1,42 @@
...
+ */
+abstract class AbstractEnumSQLDeclarationType extends AbstractEnumType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform): string
+    {
+        if (!$platform instanceof AbstractMySQLPlatform) {
+            throw new LogicException('ENUMs not supported on current platform');
```
ogizanagi:
```suggestion
            throw new LogicException('SQL ENUM type is not supported on the current platform');
```

src/Bridge/Doctrine/DBAL/Types/TypesDumper.php
```diff
@@ -49,8 +52,18 @@ public function getName(): string
...
-        $baseClass = AbstractEnumType::class;
-        $this->appendDefaultOnNullMethods($code, $enumClass, $defaultOnNull);
+        switch ($type) {
+            case self::TYPE_SCALAR:
+                $baseClass = AbstractEnumType::class;
+                $this->appendDefaultOnNullMethods($code, $enumClass, $defaultOnNull);
+                break;
+            case self::TYPE_ENUM:
+                $baseClass = AbstractEnumSQLDeclarationType::class;
+                $this->appendDefaultOnNullMethods($code, $enumClass, $defaultOnNull);
```
ogizanagi: Could be moved out of the switch

src/Bridge/Doctrine/Common/AbstractTypesDumper.php
```diff
@@ -35,7 +70,7 @@ private function dump(array $types): string
 
         $namespaces = [];
         foreach ($types as [$enumClass, $type, $name, $default]) {
-            $fqcn = self::getTypeClassname($enumClass, $type);
```
ogizanagi, line 38: getTypeClassname unused now?

src/Bridge/Symfony/Bundle/DependencyInjection/Configuration.php
```diff
@@ -71,8 +75,16 @@ private function addDoctrineDbalSection(ArrayNodeDefinition $rootNode)
                         ->end()
                         ->enumNode('type')
                             ->values(DBALTypesDumper::TYPES)
+                            ->info(<<<TXT
+Which column definition to use and the way the enumeration values are stored in the database:
+- scalar: VARCHAR/INT based on BackedEnum type
+- enum: ENUM(...values) - values are strings based on BackedEnum type (Your platform must support it)
```
ogizanagi:
```suggestion
- enum: ENUM(...values) as strings based on BackedEnum type (Your platform must support it)
```

src/Bridge/Symfony/Bundle/DependencyInjection/ElaoEnumExtension.php
```diff
@@ -113,9 +114,28 @@ private function prependDoctrineOdmConfig(array $config, ContainerBuilder $conta
...
+
+    private function resolveDbalType(array $config, bool $useEnumSQLDeclaration): string
+    {
+        $type = $config['type'];
+        $class = $config['class'];
+
+        $defaultStringType = $useEnumSQLDeclaration ? DBALTypesDumper::TYPE_ENUM : DBALTypesDumper::TYPE_SCALAR;
+
+        if (null === $type) {
+            $type = is_a($class, FlagBag::class, true) ? DBALTypesDumper::TYPE_SCALAR : $defaultStringType;
```
ogizanagi: FlagBag should not be handled here, it'll have its dedicated type.
Let's change this method to use a `getDefaultType`

tests/Fixtures/Integration/Symfony/config/config.yml
```diff
@@ -31,6 +31,9 @@ elao_enum:
         types:
             App\Enum\Suit:
                 default: !php/const App\Enum\Suit::Spades
+            #suit_sql_enum:
```
ogizanagi: Let's remove until we add a MySQL integration test case (CI only)

tests/Fixtures/Integration/Symfony/src/Entity/CardSQLEnum.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace App\Entity;
+
+use App\Enum\Suit;
+use Doctrine\ORM\Mapping as ORM;
+
+// #[ORM\Entity]
+// #[ORM\Table(name: 'cards_sql_enum')]
+class CardSQLEnum
```
ogizanagi: Same here

tests/Integration/Bridge/Doctrine/DBAL/Type/EnumTypeTest.php
```diff
@@ -53,6 +53,20 @@ public function testEnumType(): void
         self::assertSame(Suit::Hearts, $card->getSuit());
     }
 
+    public function testEnumSQLType(): void
```
ogizanagi: Will be moved into a dedicated MySQL integration test case

tests/Unit/Bridge/Doctrine/Common/Types/BaseTypesDumperTest.php
```diff
@@ -55,4 +55,18 @@ private static function updateSnapshot(string $snapshotPath, string $content): v
             file_put_contents($snapshotPath, rtrim($content) . "\n");
         }
     }
+
+    public function testPascalCaseConvertor()
+    {
+        $testCases = [
+            'Foo\\Bar\\Baz' => 'FooBarBaz',
```
ogizanagi, line 62: We should keep original namespace (and use a `@dataProvider`)

tests/Unit/Bridge/Doctrine/Common/Types/BaseTypesDumperTest.php
```diff
@@ -55,4 +55,18 @@ private static function updateSnapshot(string $snapshotPath, string $content): v
...
+            'Foo::Bar.baz' => 'FooBarBaz',
```
ogizanagi, line 63: What use-cases are these test cases covering?

tests/Unit/Bridge/Symfony/Bundle/DependencyInjection/ConfigurationTest.php
```diff
@@ -161,6 +163,7 @@ private function getDefaultConfig(): array
     {
         return [
             'doctrine' => [
+                'enum_sql_declaration' => false,
```
ogizanagi, line 192: Should add a test with true

tests/Unit/Bridge/Doctrine/DBAL/Types/TypesDumperTest.php
```diff
@@ -26,11 +26,11 @@ protected function getSnapshotPath(): string
...
-            ['Foo\Bar\Baz', 'single', 'Foo\Bar\Baz'],
-            ['Foo\Bar\BazWithDefault', 'single', 'baz_with_default', 'foo'],
-            ['Foo\Bar\Qux', 'single', 'Foo\Bar\Qux'],
-            ['Foo\Baz\Foo', 'single', 'foo'],
-            ['Foo\Baz\FooWithDefault', 'single', 'foo_with_default', 3],
+            ['Foo\Bar\Baz', 'scalar', 'Foo\Bar\Baz'],
+            ['Foo\Bar\BazWithDefault', 'scalar', 'baz_with_default', 'foo'],
+            ['Foo\Bar\Qux', 'scalar', 'Foo\Bar\Qux'],
+            ['Foo\Baz\Foo', 'scalar', 'foo'],
+            ['Foo\Baz\FooWithDefault', 'scalar', 'foo_with_default', 3],
```
ogizanagi, line 33: A test case with an enum SQL declaration is missing

tests/Unit/Bridge/Symfony/Bundle/DependencyInjection/ElaoEnumExtensionTest.php
```diff
@@ -37,9 +37,9 @@ public function testDoctrineTypes(): void
         $container = $this->createContainerFromFile('doctrine_types');
 
         self::assertEquals([
-            [Suit::class, 'single', 'suit', null],
-            [Permissions::class, 'single', 'permissions', null],
-            [RequestStatus::class, 'single', 'request_status', 200],
+            [Suit::class, 'scalar', 'suit', null],
+            [Permissions::class, 'scalar', 'permissions', null],
+            [RequestStatus::class, 'scalar', 'request_status', 200],
```
ogizanagi, line 42: A test case with `enum` is missing

benji07 (approved):

src/Bridge/Symfony/Bundle/DependencyInjection/ElaoEnumExtension.php
michnovka: I do not understand what you mean here

tests/Unit/Bridge/Doctrine/Common/Types/BaseTypesDumperTest.php
michnovka: This test is just testing the "pascalization" of various strings. The actual algorithm that decides the name acts like you say (keeps namespace) but this is tested elsewhere by comparing the dumped config with some stub saved in tests folder

tests/Unit/Bridge/Doctrine/Common/Types/BaseTypesDumperTest.php
michnovka: I want to test that All special chars are omitted and any small case letter following a nonalphanumeric char is converted to upper case

michnovka:
@ogizanagi I went over your comments and resolved most things. I also added some more tests which led me to find some bugs. All fixed.

I have created a MySQL test class. It can be enabled by changing `phpunit.xml.dist` to use MySQL instead of SQLite.

I still dont get your feedback on `FlagBag` inside `ElaoEnumExtension::resolveDbalType()`. The reason that check is there is to prevent using ENUM at any cost if FlagBag is used.

Please review once again and let me know if the way I enable MySQL tests is ok, or you wish to use some different approach. Right now I am checking `$_ENV['DOCTRINE_DBAL_URL']` for `pdo-mysql://`

michnovka:
I have a failing test on Windows due to DBAL URL I moved into `phpunit.xml.dist`

```xml
    <env name="DOCTRINE_DBAL_URL" value="sqlite:////%kernel.cache_dir%/db.sqlite" />
```

Anybody has windows who can test and tell me what it should be to work there? Thanks!

michnovka:
@ogizanagi I tried to fix it, can you re-run tests please?

michnovka:
That didnt work well. lets see if this version works, seems I had an extra `/` in the first revision.

ogizanagi:
All green now ✅ 
Thank you @michnovka for your work and investigating the issues. I'll probably make some more minor tweaks to this PR in the next days, so I'm keeping it open for now as a reminder.

michnovka:
I think it would be a good idea to set up MySQL (or MariaDB) workflow test at least for Symfony 5.4 with PHP 8.1 on Linux

ogizanagi:
That's part of the plan indeed :)

michnovka:
I finally understand the `FlagBag` issue. I can work to make `FlagBag` type as part of this PR, what do you think?

I also thought it might be nice to have an actual Symfony Form type of `FlagBag`, which would be a `select` with `multi=true`

ogizanagi:
Both of these features are listed in https://github.com/Elao/PhpEnums/issues/124 indeed.
If you'd like to work on it, that's great, but please open different PRs :)

michnovka:
Well given that I need lot of code from this PR and you plan to modify it
yet, I think Ill wait for the final merge

ogizanagi:
@michnovka : What do you mean? https://github.com/Elao/PhpEnums/pull/182 was merged

michnovka:
strange, I sent that email on April 15th :)
Screenshot 2022-10-11 at 09-58-17 Re Elao_PhpEnums Add SQL Enum support (PR #179) - michnovka@gmail com - Gmail

PR #181, opened 2022-04-15
Add FlagBag::fromAll to instantiate a bag with all of its flags

### LINKED ISSUE #180: NewFeature: Add FlagBag::ALL
Similar to `FlagBag::NONE = 0` wouldn't it be good to have `FlagBag::ALL = -1` ?

I would actually make a new enum:

```php
enum FlagBagValue: int
{
    case None = 0;
    
    case All = -1;
}
```

Then allow it in both `__constructor` and `from` and other functions accepting bits. If `FlagBagValue::All` is used, it should then analyze the underlying enum (as maybe some bit is missing in enum declaration) and treat the case equally as if every single option from the enum was passed.

This analysis could be cached in some static field of `FlagBag`.

michnovka:
Fixes https://github.com/Elao/PhpEnums/issues/180

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,8 @@ $permissions = FlagBag::from(Permissions::Execute, Permissions::Write, Permissio
 $permissions = new FlagBag(Permissions::class, 7); 
 // where 7 is the "encoded" bits value for:
 Permissions::Execute->value | Permissions::Write->value | Permissions::Read->value // 7
+// or:
+$permissions = FlagBag::fromAll(Permissions::class); // which initiates bag with all possible values
 
 $permissions = $permissions->withoutFlags(Permissions::Execute); // Returns an instance without "execute" flag
 
--- a/src/FlagBag.php
+++ b/src/FlagBag.php
@@ -34,7 +34,7 @@ class FlagBag
     private string $type;
 
     /**
-     * @param class-string<\BackedEnum>|\BackedEnum $enumOrType
+     * @param class-string<\BackedEnum> $enumType
      */
     public function __construct(string $enumType, int ...$bits)
     {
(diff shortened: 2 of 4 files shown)


michnovka:
@ogizanagi done, please merge

src/FlagBag.php
```diff
@@ -34,7 +34,9 @@ class FlagBag
     private string $type;
 
     /**
-     * @param class-string<\BackedEnum>|\BackedEnum $enumOrType
+     * @param class-string<\BackedEnum> $enumType
+     *
+     * @throws \ReflectionException
```
ogizanagi: Such errors are like logic exceptions: it should lead to a fix in the code and are never meant to be caught, so it's not useful to document it in the docblocks.

README.md
```diff
@@ -255,6 +255,11 @@ $permissions->hasFlags(Permissions::Write); // False
 Hence, using `FlagBag::getValue()` you can get an encoded value for any combination of flags from your enum, 
 and use it for storage or communication between your processes.
 
+You can initiate a "full" bag with all its flags set by calling the `fromAll` static method:
+```php
+$permissions = FlagBag::fromAll(Permissions::class); // Returns an empty bag
```
ogizanagi: Let's just add this example with the others above. For instance, after:

```php
$permissions = FlagBag::from(Permissions::Execute, Permissions::Write, Permissions::Read);
// same as:
$permissions = new FlagBag(Permissions::class, 7); 
// where 7 is the "encoded" bits value for:
Permissions::Execute->value | Permissions::Write->value | Permissions::Read->value // 7
```

add:

```php
// or, you can also initiate a bag with all its flags using:
$permissions = FlagBag::fromAll(Permissions::class);
```

ogizanagi:
Thanks @michnovka.

PR #182, opened 2022-04-16
Add SQL Enum support

michnovka:
Had some mess with branches (was working on main one, not on fork branch), so after renaming it got deleted in the old PR https://github.com/Elao/PhpEnums/pull/179

This is the same content, just different branch name is used so that I can work on other stuff before this is merged. Sorry for the mix-up, didnt know renaming branch would close existing PR, I expected itd just be renamed there too

PR #183, opened 2022-04-17
FlagBagType support

michnovka:
This is initial commit for `FlagBagType` support.

Lets discuss in here

### DIFF
--- a/src/Bridge/Symfony/Form/Type/FlagBagType.php
+++ b/src/Bridge/Symfony/Form/Type/FlagBagType.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the "elao/enum" package.
+ *
+ * Copyright (C) Elao
+ *
+ * @author Elao <contact@elao.com>
+ */
+
+namespace Elao\Enum\Bridge\Symfony\Form\Type;
+
+use Elao\Enum\Bridge\Symfony\Form\DataTransformer\FlagBagToCollectionTransformer;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Exception\InvalidConfigurationException;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @final
+ */
+class FlagBagType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        if (!$options['multiple']) {
+            throw new InvalidConfigurationException(sprintf(
+                'The "multiple" option of the "%s" form type cannot be set to false.',
+                static::class
+            ));
+        }
+
+        $builder->addModelTransformer(new FlagBagToCollectionTransformer($options['class']));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setDefault('multiple', true)
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getParent(): ?string
+    {
+        return EnumType::class;
+    }
+}
--- a/src/FlagBag.php
+++ b/src/FlagBag.php
@@ -38,13 +38,7 @@ class FlagBag
      */
     public function __construct(string $enumType, int ...$bits)
     {
-        if (!is_a($enumType, \BackedEnum::class, true)) {
-            throw new InvalidArgumentException(sprintf('"%s" is not a backed enum', $enumType));
-        }
-
-        if ('int' !== (string) (new \ReflectionEnum($enumType))->getBackingType()) {
-            throw new InvalidArgumentException(sprintf('"%s" is not an int backed enum', $enumType));
-        }
+        self::checkIntBackedEnumType($enumType);
 
         $this->type = $enumType;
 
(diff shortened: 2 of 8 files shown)


michnovka:
I have ported also `BitmaskToBitFlagsValueTransformer`, but I dont think this is going to be used now. In fact the whole `as_value` part makes little to no sense for me under the new implementation. One of the issues is that the new 2.x `EnumType` extends the Symfony's own `EnumType` and not the `ChoiceType` which was the case with 1.x version. So if we use `DataTransformer` that spits out `int` and not `BackedEnum[]`, then the Symfony `EnumType` does not handle it and fails.

I have made `BitmaskToBitFlagsValueTransformerTest` so I think we can keep the DataTransformer in the package, maybe somebody will find it useful. But Id remove the `as_value` option from the `FlagBagType` and would only allow it to work with the `FlagBagToCollectionTransformer` (refactored version of `SingleToCollectionFlagEnumTransformer`)

WDYT @ogizanagi ?

ogizanagi:
> I have ported also BitmaskToBitFlagsValueTransformer, but I dont think this is going to be used now. In fact the whole as_value part makes little to no sense for me under the new implementation.

I agree, let's remove this part until it is proven useful. 
I don't want to strictly port every 1.x features as is, only what makes sense regarding the new PHP 8.1 native enums implementation. Also we don't need everything to land in 2.0.0, just what can impact the design and main interfaces.

michnovka:
> Also we don't need everything to land in 2.0.0, just what can impact the design and main interfaces.

Yes, but I would like to include SQL enum, this PR and also FlagBag DBAL type for sure

michnovka:
This is ready for merge @ogizanagi

michnovka:
One important note: I have made the `FlagBag::getBitmask()` public. I think its a useful tool to be used with FlagBags and dont see any reason why it should stay protected.

benji07 (approved):

src/Bridge/Symfony/Form/DataTransformer/BitmaskToBitFlagsValueTransformer.php
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @author Elao <contact@elao.com>
+ */
+
+namespace Elao\Enum\Bridge\Symfony\Form\DataTransformer;
+
+use Elao\Enum\FlagBag;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+class BitmaskToBitFlagsValueTransformer extends AbstractFlagBagTransformer
```
ogizanagi: Let's remove it completely

src/FlagBag.php
```diff
@@ -57,6 +51,24 @@ public function __construct(string $enumType, int ...$bits)
         $this->bits = static::decodeBits($bits);
     }
 
+    /**
+     * @param class-string $enumType
+     *
+     * @throws InvalidArgumentException
```
ogizanagi: That's a logic exception, let's not document it

src/FlagBag.php
```diff
@@ -57,6 +51,24 @@ public function __construct(string $enumType, int ...$bits)
...
+     *
+     * @return void
```
ogizanagi: Let's use `void` return type on the signature directly

src/FlagBag.php
```diff
@@ -57,6 +51,24 @@ public function __construct(string $enumType, int ...$bits)
...
+     */
+    public static function checkIntBackedEnumType(string $enumType)
```
ogizanagi: Let's keep this private for now

```suggestion
    private static function checkIntBackedEnumType(string $enumType): void
```

src/Bridge/Symfony/Form/Type/FlagBagType.php
```diff
@@ -0,0 +1,71 @@
...
+            ->setDefault('multiple', true)
+            ->setAllowedValues('class', static function (string $enumType): bool {
+                if (!is_a($enumType, \BackedEnum::class, true)) {
+                    return false;
+                }
+
+                try {
+                    // this throws if some value is not power of 2
+                    FlagBag::getBitmask($enumType);
+                } catch (LogicException $e) {
```
ogizanagi: I think we should not even care here, the exception will be thrown with a clear enough message for the user to fix it. Let's remove this `setAllowedValues`.

tests/Unit/Bridge/Symfony/Form/DataTransformer/FlagBagToCollectionTransformerTest.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->expectExceptionMessage('Expected FlagBag instance of "Elao\Enum\Tests\Fixtures\Enum\Permissions" values. Got FlagBag instance of "Elao\Enum\Tests\Fixtures\Enum\PermissionsMissingBit" values.');
+
+        $transformer = new FlagBagToCollectionTransformer(Permissions::class);
+
+        $transformer->transform(new FlagBag(PermissionsMissingBit::class, FlagBag::NONE));
+    }
+
+    public function provideCollectionToFlagBag(): \Traversable
+    {
+        yield from array_map('array_reverse', iterator_to_array($this->provideFlagBagToCollection()));
```
ogizanagi:
```suggestion
        yield from array_map(array_reverse(...), iterator_to_array($this->provideFlagBagToCollection()));
```

michnovka:
@ogizanagi changes committed, ready for merge

ogizanagi:
Thanks @michnovka for working on this feature, this is much appreciated.

michnovka:
@ogizanagi np, I use this lib so I like to contribute. I will see if I can work on FlagBag DBAL type when https://github.com/Elao/PhpEnums/pull/182 is merged

PR #184, opened 2022-04-22
Add ExtrasInterface

michnovka:
This is useful as checking for trait support is quite hard and inefficient in PHP.

This allows you to use `is_a($some_enum, Elao\Enum\ExtrasInterface::class, true)` and `$some_enum instanceof Elao\Enum\ExtrasInterface::class` for enums that should support the `getExtra` method

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,8 @@ The `EnumCase` attributes also provides you a way to configure some extra attrib
 ```php
 namespace App\Enum;
 
-use Elao\Enum\ExtrasInterface;use Elao\Enum\ReadableEnumInterface;
+use Elao\Enum\ExtrasInterface;
+use Elao\Enum\ReadableEnumInterface;
 use Elao\Enum\ExtrasTrait;
 use Elao\Enum\Attribute\EnumCase;
 
--- a/README.md
+++ b/README.md
@@ -133,11 +133,11 @@ The `EnumCase` attributes also provides you a way to configure some extra attrib
 ```php
 namespace App\Enum;
 
-use Elao\Enum\ReadableEnumInterface;
+use Elao\Enum\ExtrasInterface;use Elao\Enum\ReadableEnumInterface;
 use Elao\Enum\ExtrasTrait;
 use Elao\Enum\Attribute\EnumCase;
 
-enum Suit implements ReadableEnumInterface
+enum Suit implements ExtrasInterface
 {
     use ExtrasTrait;
 
@@ -167,7 +167,9 @@ Suit::Spades->getExtra('missing-key', true); // throws
 or create your own interfaces/traits:
 
 ```php
-interface RenderableEnumInterface 
+use Elao\Enum\ExtrasInterface;
+
+interface RenderableEnumInterface extends ExtrasInterface
 {
     public function getColor(): string;
     public function getIcon(): string;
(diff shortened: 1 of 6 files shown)


ogizanagi:
Do you have a use-case for this interface? I'd like to avoid adding new interfaces for no reasons.
The extras feature was designed with userland usages in mind, for which you should define your own interfaces where it makes sense, as described in https://github.com/Elao/PhpEnums#extra-values last sample. The trait only is a convenient helper.

michnovka:
Well I do have a specific use-case in my project, where i want to check if there any extras set on enum cases. Yes, I can make my own interfaces (which I am doing now). This is totally your call, I just have a convention in my projects that for every trait, I make an interface, as it makes it so much easier to check whether that trait is supported.

A very specific case imo is when you want to use ONLY extras, but not `ReadableEnumInterface` methods. Both `ExtrasTrait` and `ReadableEnumTrait` make use of `EnumCaseAttributesTrait`. In my code then, when I have 

```php
enum SuitWithExtras
{
    use ExtrasTrait;

    #[EnumCase(extras: ['icon' => 'fa-heart', 'color' => 'red', 'only-for-hearts' => 'value'])]
    case Hearts;

    #[EnumCase(extras: ['icon' => 'fa-diamond', 'color' => 'red'])]
    case Diamonds;

    #[EnumCase(extras: ['icon' => 'fa-club', 'color' => 'black'])]
    case Clubs;

    #[EnumCase(extras: ['icon' => 'fa-spade', 'color' => 'black'])]
    case Spades;
}
```

its hard to check whether it actually is a `EnumCase` enum.

Using

```php
enum SuitWithExtras implements ExtrasInterface
```

Makes this easier without the need to include `ReadableEnumTrait` trait because of implementing `ReadableEnumInterface`

I went even further against your wish to not add more interfaces :D

I added a EnumCaseInterface, which is totally blank, BUT it allows me to mark any enum as a one supporting EnumCase attributes.

michnovka:
The `ExtrasInterface` and `ReadableEnumInterface` both extend the empty `EnumCaseInterface` so implementing either in your enum will give you a positive in `instanceof` and `is_a` check.

This makes it very convenient for checks whether the enum is just a "basic" `UnitEnum` or one that implements some of the features from this package.

ogizanagi:
> Yes, I can make my own interfaces (which I am doing now). This is totally your call, I just have a convention in my projects that for every trait, I make an interface, as it makes it so much easier to check whether that trait is supported.

That's nice, but only useful when there is an actual consumer to such interfaces. For now, we don't have such in the core library, nor it would be useful for third-party integrations (until proven) and the "extras" feature is too broad and generic ; it was only designed as a help to declare and access custom attributes in userland, where more scoped and meaningful interfaces & features can be built on top of it.

> I added a EnumCaseInterface, which is totally blank, BUT it allows me to mark any enum as a one supporting EnumCase attributes.

> The ExtrasInterface and ReadableEnumInterface both extend the empty EnumCaseInterface so implementing either in your enum will give you a positive in instanceof and is_a check.
> This makes it very convenient for checks whether the enum is just a "basic" UnitEnum or one that implements some of the features from this package.

We do not need to detect such a thing for now. 
So, thank you for taking the time to open this PR, but I won't add these interfaces until then.

michnovka:
sure, no hard feelings. My main point which I maybe did not communicate well enough was that now there is no way to distinguish between regular PHP enums and enums that have `EnumCase`. You can do it if you implement `ReadableEnumInterface`. But the current code allows you to implement `EnumCase` without readables, only with extras, and in such case there is no interface to hint that `EnumCase` is expected in the code

PR #189, opened 2022-05-31
[Docs] Explain how to use enums with Doctrine ORM QueryBuilder

pluk77:
Add section to explain how to use enum objects with QueryBuilder

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Table of Contents
         * [Using the Doctrine Bundle with Symfony](#using-the-doctrine-bundle-with-symfony)
       * [Mapping](#mapping)
       * [Default value on null](#default-value-on-null)
+      * [Troubleshooting](#troubleshooting)
     * [Doctrine ODM](#doctrine-odm)
     * [Symfony HttpKernel component](#symfony-httpkernel-component)
     * [Symfony Serializer component](#symfony-serializer-component)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -615,6 +616,29 @@ elao_enum:
...
+}
+
+```
+
+```php
+$em->createQueryBuilder()
+  ->select('c')
+  ->from('Card', 'c')
+  ->where('c.suitType = :param1')
+  ->setParameter('param1', SuitEnum::get(SuitEnum::SPADES), 'suit');
```
ogizanagi: Well, there is a simpler alternative: always use the backed value:

```suggestion
  ->setParameter('param1', SuitEnum::SPADES->value);
```

IonBazan: This is v1 so it's either:
```suggestion
  ->setParameter('param1', SuitEnum::SPADES);
```

or:
```suggestion
  ->setParameter('param1', SuitEnum::SPADES(), 'suit');
```

(using the Doctrine type to convert it to right value).

ogizanagi: Oh right, I missed this was targeting the 1.x branch.

pluk77: Maybe the example is not the best. I agree that with static values, using the actual constant is the most strait forward and probably preferred choice.

Using a return value which is either null or an enum will not always work correctly if the 3rd parameter is not set. This is what my PR tries to explain. Any suggestions on how to best put this in the docs?

The problem is that if the return value of the function is an enum, the query does not fail and no errors are given. It will use __toString() and (generally) use the incorrect value in the actual query.

```php
$em->createQueryBuilder()
  ->select('c')
  ->from('Card', 'c')
  ->where('c.suitType = :param1')
  ->setParameter('param1', $filter->getSuitType(), 'suit');
```

ogizanagi: What about:

```php
$em->createQueryBuilder()
  ->select('c')
  ->from('Card', 'c')
  ->where('c.suitType = :param1')
  ->setParameter('param1', $filter->getSuitType()?->getValue());
  // or, PHP < 8
  ->setParameter('param1', $filter->getSuitType() ? $filter->getSuitType()->getValue() : null);
```

? But you can include your third syntax as well and let the users choose what they prefer.

ogizanagi:
Thanks @pluk77.

pluk77:
> Thanks @pluk77.

Glad I could be of assistance.

PR #193, opened 2022-06-20
Fix PhpStan issue with FlagBag missing template on static methods

laurentmuller:


ogizanagi:
Thank you @laurentmuller.

PR #201, opened 2022-11-08
[Readable] Allow to configure a common suffix/prefix & use the enum case value/name as default label

ogizanagi:
Eases the declaration of readable enum translations keys with common suffix/prefix:

```php
#[ReadableEnum(prefix: 'suit.')]
enum Suit: string implements ReadableEnumInterface
{
    use ReadableEnumTrait;

    case Hearts = '♥︎';
    case Diamonds = '♦︎';
    case Clubs = '♣︎';
    case Spades = '︎♠︎';
}

Suit::Hearts->getReadable(); // returns 'suit.Hearts'
Suit::Clubs->getReadable(); // returns 'suit.Clubs'
```

[Docs](https://github.com/Elao/PhpEnums/blob/readable-auto/README.md#configure-suffixprefix--default-value)

## Questions

- Should we actually default on the `name` rather than the value by default when using this attribute?
- Name this attribute `ConfigureReadableEnum`? `AutoReadableEnum`?
- …

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,21 +12,15 @@ _Provides additional, opinionated features to the [PHP 8.1+ native enums](https:
 as specific integrations with frameworks and libraries._
 
 ```php
+#[ReadableEnum(prefix: 'suit.')]
 enum Suit: string implements ReadableEnumInterface
 {
     use ReadableEnumTrait;
 
-    #[EnumCase('suit.hearts')]
-    case Hearts = 'H';
-
-    #[EnumCase('suit.diamonds')]
-    case Diamonds = 'D';
-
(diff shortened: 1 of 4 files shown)


benji07 (approved):

IonBazan (approved):
Looks nice and much easier to use with enums having many cases

chalasr:
> Should we actually default on the name rather than the value by default when using this attribute?

👍 I think so

> Name this attribute ConfigureReadableEnum? AutoReadableEnum?

Not necessarily IMHO it's simpler as-is

benji07 (approved):

PR #216, opened 2023-02-20
allow int enum to be normalized as int

maximecolin:


src/Bridge/Symfony/Serializer/Normalizer/EnumNormalizer.php
```diff
@@ -27,8 +27,10 @@ class EnumNormalizer implements NormalizerInterface, DenormalizerInterface
      * {@inheritdoc}
      *
      * @param EnumInterface $object
+     *
+     * @return string|int
```
benji07:
```suggestion
     * @return mixed
```

Peut être voir si on peux mettre des générics phpstan ici ( en mettant le param EnumInterface<T> et return T)

ogizanagi: I don't think it'll works here, since we have no way to indicate the template type.
It works when extending a base class defining a template. I'm not sure it's valid as is.

At least it misses `@template T of int|string` ?

ogizanagi:
Thanks Maxime.

PR #219, opened 2023-03-22
Add phpstan

benji07:
Tests fail because of a deprecation inside of Doctrine ODM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,12 +22,30 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: '8.1'
+          coverage: none
+
+      - name: 'Require Doctrine MongoDB dependencies'
+        run: composer require --no-update --ignore-platform-reqs --dev --no-interaction --ansi "doctrine/mongodb-odm:^2.4" "doctrine/mongodb-odm-bundle:^4.5.1"
 
       - name: 'Install dependencies'
-        run: make php-cs-fixer.phar
+        id: deps
+        run: |
+          echo "::group::Install project deps"
+            composer update --prefer-dist --ignore-platform-reqs --ansi
+          echo "::endgroup::"
 
-      - name: 'Check style'
-        run: ./php-cs-fixer.phar fix --dry-run --no-interaction --diff
+          echo "::group::Install PHPUnit"
+            vendor/bin/simple-phpunit install
+          echo "::endgroup::"
+        env:
+          SYMFONY_REQUIRE: "6.2*dev"
+
+      - name: 'Lint - php-cs-fixer'
+        run: make lint.php-cs-fixer@integration
+
+      - name: 'Lint - phpstan'
+        if: always() && steps.deps.outcome == 'success'
+        run: make lint.phpstan@integration
 
   test:
     name: ${{ matrix.name }}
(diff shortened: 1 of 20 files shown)


.github/workflows/ci.yml
```diff
@@ -29,6 +29,38 @@ jobs:
...
+    timeout-minutes: 5
+
+    steps:
+      - name: 'Checkout'
+        uses: actions/checkout@v3
+
+      - name: 'Setup PHP'
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.1'
```
ogizanagi: Let's disable Xdebug here: 

image

https://github.com/shivammathur/setup-php#disable-coverage

(perhaps as well for other workflows)

.github/workflows/ci.yml
```diff
@@ -29,6 +29,38 @@ jobs:
...
+          echo "::endgroup::"
+
+          echo "::group::Install PHPUnit"
+            vendor/bin/simple-phpunit install
+          echo "::endgroup::"
+        env:
+          SYMFONY_REQUIRE: "6.2*dev"
+
+      - name: 'Run PHPStan'
+        run: vendor/bin/phpstan
```
ogizanagi: Let's update the `Makefile` as well so we can easily run Phpstan when using `make lint` or `make lint.fix` ?

.github/workflows/ci.yml
```diff
@@ -29,6 +29,38 @@ jobs:
       - name: 'Check style'
         run: ./php-cs-fixer.phar fix --dry-run --no-interaction --diff
 
+  phpstan:
```
ogizanagi: What about doing this in the `lint` job? Any value to have this in a dedicated job?

There is still a way to run Phpstan even if the `Check style` step is failing with `if: always() && […]`. See https://github.com/Elao/elao_/blob/292aa8b09ae29faadb3bbd8eaa480877eb6fc473/.github/workflows/tests.yaml#L75-L103 for instance

src/FlagBag.php
```diff
@@ -68,19 +70,17 @@ private static function checkIntBackedEnumType(string $enumType): void
...
     public static function fromAll(string $enumType): FlagBag
     {
         return new FlagBag($enumType, static::getBitmask($enumType));
     }
 
     /**
-     * @template T of \BackedEnum
-     *
-     * @param class-string<T>|T $enumOrType
+     * @param class-string<\BackedEnum>|\BackedEnum $enumOrType
```
ogizanagi: Wasn't the previous code more accurate for this? Can't you keep `class-string<T>|T $enumOrType` ?

src/FlagBag.php
```diff
@@ -103,9 +103,7 @@ public static function from(string|\BackedEnum $enumOrType, \BackedEnum ...$flag
     }
 
     /**
-     * @template T of \BackedEnum
-     *
-     * @param class-string<T> $enumType
+     * @param class-string<\BackedEnum> $enumType
```
ogizanagi: Same

ogizanagi (approved):

.github/workflows/ci.yml
ogizanagi: I meant for running make lint or make lint.fix on your host, not on the CI 😄

ogizanagi:
Thanks @benji07.

Issue #221, opened 2023-04-04
Implement TranslatableInterface for the ReadableEnumInterface


laurentmuller:
Because the `ReadableEnumInterface` can be translated within the `TranslatorInterface`, it will be nice to implements also the `TranslatableInterface` to enums.

To translate readable enums is just matter to have a call as follow:
```php

public function trans(TranslatorInterface $translator, string $locale = null): string
{
    return $translator->trans($this->getReadable(), [], $locale);
}
```
We then use in **PHP**:
```php

NotifyType::EMAIL->trans($this->translator)

```
Or in **Twig**:
```twig
{{ supplier.notifyType|trans }}
```

We can maybe create a trait for that:
```php
use Elao\Enum\ReadableEnumTrait;
use Symfony\Contracts\Translation\TranslatorInterface;

trait TranslatableEnumTrait {

    use ReadableEnumTrait;

    public function trans(TranslatorInterface $translator, string $locale = null): string
    {
        return $translator->trans($this->getReadable(), [], $locale);
    }
}
```

And in a enums:
```php
use Elao\Enum\ReadableEnumInterface;
use Symfony\Contracts\Translation\TranslatableInterface;

enum NotifyType: int implements ReadableEnumInterface, TranslatableInterface
{
    use TranslatableEnumTrait ;

    #[EnumCase('notify.email')]
    case EMAIL = 0;
    #[EnumCase('notify.webhook')]
    case WEBHOOK = 1;
}
```

PR #224, opened 2023-07-05
Implementation of the translatable interface.

laurentmuller:
This is the PR for the feature #221.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ enum Suit: string implements ReadableEnumInterface
   - [Symfony Forms](#symfony-form)
   - [Symfony Controller Argument Resolver](#symfony-httpkernel)
   - Symfony VarDumper
+  - [Symfony Translation](#symfony-translation)
   - [Doctrine ORM](#doctrine)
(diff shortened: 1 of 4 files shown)


ogizanagi (approved):
I like this. Could you please add a section describing this and the usage in the README.md file ?

laurentmuller:
Thank You. I will update the `README.md` file.

ogizanagi (approved):

ogizanagi:
Thank you @laurentmuller.

laurentmuller:
You'r welcome.

laurentmuller:
Any release ?

IonBazan:
I think we should add some tests for this 🤔

norkunas:
Since https://github.com/symfony/symfony/pull/50931 if `TranslatableInterface` would be used instead of `TranslatableEnumInterface` there would be no need to have a `Elao\Enum\Bridge\Symfony\Form\Type\EnumType` and directly use one from Symfony.

edit: missed that `TranslatableEnumInterface` extends `TranslatableInterface`. Just don't see any point from that interface

Issue #192, opened 2022-06-17
Pslam errors with FlagBag


laurentmuller:
I have some errors when I use the `FlagBag` class and test within Psalm. 

Example:

```php
namespace App\Enum;

enum Permissions: int
{
    case Execute = 1 << 0;
    case Write = 1 << 1;
    case Read = 1 << 2;
}
```
I then use this enumeration within FlagBag:

```php
$flagBag = FlagBag::from(Permissions::class);
```

Pslam report this error:

```
ERROR: InvalidArgument - Argument 1 of Elao\Enum\FlagBag::from expects Elao\Enum\T|class-string<Elao\Enum\T>, App\Enum\Permissions::class provided (see https://psalm.dev/004)
        FlagBag::from(Permissions::class);
```
We have annoted the Permissions class ?

ogizanagi:
does adding:

```php
 * @template T of \BackedEnum
```

 to the `FlagBag::from` method solves this?

laurentmuller:
Yes, with this change, Pslam does not more complain about invalid argument.

ogizanagi:
PR welcomed then :)

ogizanagi:
Fixed in #193

laurentmuller:
Hello,

I open this error again because Pslam complains about the following code:

```php
private function getDefaultPermissions(): FlagBag
{
    return FlagBag::from(Permissions::Execute , Permissions::Write, Permissions::Read);
}
```

`Argument 2 of Elao\Enum\FlagBag::from expects Elao\Enum\T, but enum(App\Enum\Permissions::Execute) provided (see https://psalm.dev/004)`

`Argument 3 of Elao\Enum\FlagBag::from expects Elao\Enum\T, but enum(App\Enum\Permissions::SHOW) provided (see https://psalm.dev/004)`

I have modified the `from` function with this annotation:

```php
/**
  * @template E of \BackedEnum
  * @param class-string<E>|E $enumOrType
  * @param E ...$flags
  * @return FlagBag<E>
  */
public static function from(string|\BackedEnum $enumOrType, \BackedEnum ...$flags): static
{
    //...
}
```
And it work without error.

ogizanagi:
It seems the suggested code triggers an issue with PHPStan : 

```shell
 ------ -----------------------------------------------------------------------------------------------------------------------------
  Line   FlagBag.php
 ------ -----------------------------------------------------------------------------------------------------------------------------
[...]

PR #226, opened 2023-09-27
Updated FlagBag.

laurentmuller:
This is the correction for issue #192.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ enum Suit: string implements ReadableEnumInterface
   - [Symfony Forms](#symfony-form)
   - [Symfony Controller Argument Resolver](#symfony-httpkernel)
   - Symfony VarDumper
+  - [Symfony Translation](#symfony-translation)
   - [Doctrine ORM](#doctrine)
   - [Doctrine ODM](#doctrine-odm)
   - [Faker](#faker)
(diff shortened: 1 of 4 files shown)


src/FlagBag.php
```diff
@@ -70,17 +70,26 @@ private static function checkIntBackedEnumType(string $enumType): void
...
     /**
-     * @param class-string<T>|T $enumOrType
+     * @template E of \BackedEnum
+     *
+     * @param class-string<E>|E $enumOrType
+     * @param E                 ...$flags
+     *
+     * @return FlagBag<E>
+     *
+     * @phpstan-return FlagBag<T>
```
ogizanagi: Why would this one be required here while not in `fromAll` ?

```suggestion
```

laurentmuller: No idea.

But if this is not present, PHP Stan compain about this:

```
------ ----------------------------------------------------------------------------------------------------------------------------------------------
  Line   FlagBag.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------
  93     PHPDoc tag @return with type Elao\Enum\FlagBag<E of BackedEnum> is incompatible with native type static(Elao\Enum\FlagBag<T of BackedEnum>).
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------
```

laurentmuller: One more time, for me what is working is:

```php
/**
 * @param class-string<T>|T $enumOrType
 *
 * @phpstan-return FlagBag<T>
 *
 * @psalm-template E of \BackedEnum
 * @psalm-param E ...$flags
 * @psalm-return FlagBag<E>
 */
public static function from(string|\BackedEnum $enumOrType, \BackedEnum ...$flags): static
{
	// ...
}
```

PR #227, opened 2024-01-03
Bump dependencies. test against symfony 7

michnovka:
Update tests and check against versions that are not EOL:

https://endoflife.date/symfony

### DIFF
--- a/src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
+++ b/src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
@@ -95,14 +95,17 @@ abstract class AbstractEnumType extends Type
      */
     public function getSQLDeclaration(array $column, AbstractPlatform $platform): string
     {
-        return $this->isIntBackedEnum()
-            ? $platform->getIntegerTypeDeclarationSQL($column)
-            : (
-                method_exists($platform, 'getStringTypeDeclarationSQL') ?
-                $platform->getStringTypeDeclarationSQL($column) :
-                $platform->getVarcharTypeDeclarationSQL($column)
-            )
-        ;
+        if ($this->isIntBackedEnum()) {
+            return $platform->getIntegerTypeDeclarationSQL($column);
+        }
+
+        if (empty($column['length'])) {
+            $column['length'] = 255;
+        }
+
+        return method_exists($platform, 'getStringTypeDeclarationSQL') ?
+            $platform->getStringTypeDeclarationSQL($column) :
+            $platform->getVarcharTypeDeclarationSQL($column);
     }
 
     /**
--- a/tests/Fixtures/Integration/Symfony/config/config-routing-annotation.yaml
+++ b/tests/Fixtures/Integration/Symfony/config/config-routing-annotation.yaml
@@ -1,0 +1,3 @@
+framework:
+    router:
+        resource: '%kernel.project_dir%/config/routing-annotation.yaml'
--- a/tests/Fixtures/Integration/Symfony/config/config-routing-attribute.yaml
+++ b/tests/Fixtures/Integration/Symfony/config/config-routing-attribute.yaml
@@ -1,0 +1,5 @@
+framework:
+    annotations:
+            enabled: false
+    router:
+        resource: '%kernel.project_dir%/config/routing.yaml'
--- a/tests/Fixtures/Integration/Symfony/config/doctrine-new.yaml
+++ b/tests/Fixtures/Integration/Symfony/config/doctrine-new.yaml
@@ -1,0 +1,4 @@
+doctrine:
+    orm:
+        report_fields_where_declared: true
+        enable_lazy_ghost_objects: true
--- a/tests/Fixtures/Integration/Symfony/src/ControllerAnnotation/QueryBodyBackedEnumValueResolverController.php
+++ b/tests/Fixtures/Integration/Symfony/src/ControllerAnnotation/QueryBodyBackedEnumValueResolverController.php
@@ -10,14 +10,14 @@ declare(strict_types=1);
  * @author Elao <contact@elao.com>
  */
 
-namespace App\Controller;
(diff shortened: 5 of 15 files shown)


michnovka:
I am sorry, I dont have the time nor energy to work on this.. I assumed itd be a simple fix. But there are many changes in Doctrine as well as in symfony, tests are therefore failing.

ogizanagi:
No problem, thanks @michnovka for trying. Keeping opened for reminder.

michnovka:
@ogizanagi  Have a look at the errors tho. What buffles me most is that ini_set error.

Then we have another issue - doctrine since orm 2.17. requires some config params

```yaml
        report_fields_where_declared: true
        enable_lazy_ghost_objects: true
```

and throws deprecation when not present.
But for older versions, these cannot be present at all, as it will throw and exit.

So with "lowest deps" option we need to use different config for symfony than for all the others (which default to 2.17 version as of now).

The mongodb installation on windows is totally broken, but id look into this as latest.

michnovka:
I made some progress, but we have an issue still.

There are multiple config conflicts we need to address depending on which CI job is run (which Symfony/Doctrine version)

One is the aforementioned 

```yaml
        report_fields_where_declared: true
        enable_lazy_ghost_objects: true
```

That stems from DoctrineBundle version < 2.10 cannot parse the config. And if we do bump the bundle, we still need to keep these values as FALSE for orm < 2.14, as it is not supported. So we need 2 different config files for `tests/Fixtures/Integration/Symfony/config/config.yaml`

then in Symfony 7.x annotation routing was removed, so we also need 2 versions of `tests/Fixtures/Integration/Symfony/config/routing.yaml`

I dont know how to do this, so @ogizanagi if you have an idea, or preferably if you build on my findings I will appreciate any help.

---

I dont get AT ALL why symfony 6.x has the ini_set errors.

---

And also still totally unsure why Windows mongodb install suddenly fails, I didnt make changes there.

michnovka:
Update - I followed https://github.com/symfony/symfony/issues/51600 (no idea why this is needed, but it does work) and now symfony 6.3 works well.

We still have tons of deprecations issues in 6.4 and 7.x does not work at all. We need separate configs based on the symfony version used

michnovka:
My OCD will kill me one day... Cant leave a draft PR.

Anyways, I am happy with it, please review @ogizanagi

src/Bridge/Doctrine/DBAL/Types/AbstractEnumType.php
```diff
@@ -95,14 +95,17 @@ public function convertToPHPValue($value, AbstractPlatform $platform): ?\BackedE
...
-            : (
-                method_exists($platform, 'getStringTypeDeclarationSQL') ?
+        if ($this->isIntBackedEnum()) {
+            return $platform->getIntegerTypeDeclarationSQL($column);
+        } else {
+            if (empty($column['length'])) {
+                $column['length'] = 255;
+            }
+
+            return method_exists($platform, 'getStringTypeDeclarationSQL') ?
```
benji07: You can remote the else as you do an early return in the `if` statement

```suggestion
        if ($this->isIntBackedEnum()) {
            return $platform->getIntegerTypeDeclarationSQL($column);
        }
        
        if (empty($column['length'])) {
            $column['length'] = 255;
        }

        return method_exists($platform, 'getStringTypeDeclarationSQL') ?
```

tests/Fixtures/Integration/Symfony/config/config-routing-annotation.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+    router:
+        resource: '%kernel.project_dir%/config/routing-annotation.yaml'
```
benji07: missing an extra line at the end of the file

tests/Fixtures/Integration/Symfony/config/config-routing-attribute.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+    annotations:
+            enabled: false
+    router:
+        resource: '%kernel.project_dir%/config/routing.yaml'
```
benji07: missing an extra line at the end of the file

tests/Fixtures/Integration/Symfony/src/ControllerAnnotation/QueryBodyBackedEnumValueResolverController.php
```diff
@@ -0,0 +1,89 @@
...
+{
+    use VarDumperTestTrait;
+
+    public function __construct()
+    {
+        $this->setUpVarDumper([], CliDumper::DUMP_LIGHT_ARRAY);
+    }
+
+    #[Route(path: '/from-query')]
+    public function fromQuery(
```
benji07, line 35: the return type are missing of every of the method here

tests/Fixtures/Integration/Symfony/config/doctrine-new.yaml
```diff
@@ -0,0 +1,4 @@
+doctrine:
+    orm:
+        report_fields_where_declared: true
+        enable_lazy_ghost_objects: true
```
benji07: missing an extra line at the end of the file

tests/Fixtures/Integration/Symfony/config/doctrine-old.yaml
benji07, line 1: is this file supposed to be empty ?

michnovka: yea, in the end I didnt need to put anything there, but maybe later we will want to specify some config for old doctrine. I can remove it now, but doesnt hurt to have it.

benji07 (approved):

michnovka:
do you want me to squash commits?

benji07:
@michnovka yes, please 🙏🏻

ogizanagi (approved):
Awesome work, thank you very much @michnovka 🙏🏻

PR #228, opened 2024-01-05
WIP: DBAL4 and ORM3 support

michnovka:
This PR reflects some DBAL changes in version 4.0

We will likely have to create new major version, as there is no way to keep in one version support for both DBAL3 and DBAL4.

For this purpose I have dropped Symfony 5.x support. Symfony 5.x and DBAL3 should be maintained in 2.x branch

WORK IN PROGRESS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,7 @@ jobs:
             php: '8.1'
             symfony: '6.4.*@dev'
             mongodb: true
+            mongodbnew: true
             mysql: true
             allow-unstable: true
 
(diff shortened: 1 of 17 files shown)


michnovka:
I think we should look into this now and plan what to do since ORM3 and DBAL4 is now out and in real use.

Due to the underlying changes we cannot have one major version work with both.

I propose to make new version 3.0 which will target ORM3 and DBAL4 and keep 2.x for ORM2 and DBAL3.

ogizanagi:
Thanks for notifying. Considering the few changes in this PR about doctrine, do you really think it'll require much more efforts to maintain both versions support in 2.x?

PR #235, opened 2024-10-15
Conflict broken DBAL 4 versions

### LINKED ISSUE #234: Bump DBAL 4 min version to 4.2.0
There was an issue with infinite migrations being generated when using DBAL 4 - https://github.com/doctrine/migrations/issues/1441

After a lot of discussion, there came a fix from doctrine team which works with MySQL ENUMs - https://github.com/doctrine/dbal/pull/6536

I would argue that this is a required min. version of DBAL in order for the plugin to work without issues.

Shall we bump dependency?

michnovka:
Fixes #234

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     "require-dev": {
         "ext-pdo_sqlite": "*",
         "composer/semver": "^3.2",
-        "doctrine/dbal": "^3.8|^4.0",
+        "doctrine/dbal": "^3.8|^4.2",
         "doctrine/doctrine-bundle": "^2.11",
         "doctrine/orm": "^2.20|^3.0",
         "phpstan/phpstan": "^1.10",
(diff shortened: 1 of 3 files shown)


michnovka:
Hmm, so we have 2 failed tests, separate issues:

1. the lowest deps installs `symfony/var-exporter (v4.4.11)` and the error says:
```
LogicException: Lazy ghost objects cannot be enabled because the "symfony/var-exporter" library version 6.2 or higher is not installed. Please run "composer require symfony/var-exporter:^6.2".
```
I want to investigate that.

2. PHP 8.4 fails due to PHPUnit's incompatibility with PHP8.4 Id say (there are deprecations), I think safe to ignore

michnovka:
Ok, I think this can be merged. The php 8.4 failed test cant be fixed until we use phpunit 11, and unfortunately thats not compatible with symfony phpunit bridge.
I think the future of this bridge is doomed anyways, so at one point well have to use native phpunit without bridge, but lets see. For now I am done. @ogizanagi please review

ogizanagi:
Thanks @michnovka.

PR #237, opened 2025-01-18
Updated the FlagBag templates.

laurentmuller:
This is the correction for issue https://github.com/Elao/PhpEnums/issues/192. 
It's seem that I never created a pull request for this.

### DIFF
--- a/src/FlagBag.php
+++ b/src/FlagBag.php
@@ -76,9 +76,11 @@ class FlagBag
     }
 
     /**
-     * @param class-string<T> $enumType
+     * @template E of \BackedEnum
      *
-     * @return FlagBag<T>
+     * @param class-string<E> $enumType
+     *
+     * @return FlagBag<E>
      */
     public static function fromAll(string $enumType): FlagBag
     {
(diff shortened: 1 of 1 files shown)


src/FlagBag.php
```diff
@@ -81,6 +83,14 @@ public static function fromAll(string $enumType): FlagBag
...
     /**
      * @param class-string<T>|T $enumOrType
+     *
+     * @phpstan-return FlagBag<T>
+     *
+     * @psalm-template E of \BackedEnum
+     *
+     * @psalm-param E ...$flags
+     *
+     * @psalm-return FlagBag<E>
```
ogizanagi: Why does it need another template here? Can't it be:

```suggestion
     *
     * @template E of \BackedEnum
     *
     * @param class-string<E>|E $enumOrType
     * @param E[] ...$flags
     *
     * @return FlagBag<E>
```

? The backed enum type is supposed to be the same

ogizanagi:
Thanks @laurentmuller.

src/FlagBag.php
laurentmuller: Please see https://github.com/Elao/PhpEnums/issues/192#issuecomment-1573722202

PR #243, opened 2025-08-25
[Draft] Upgrade for symfony 7.3 & php > 8.3

brouquet-company:


.gitignore
```diff
@@ -3,3 +3,4 @@
 .php-cs-fixer.cache
 php-cs-fixer.phar
 .phpunit.result.cache
+.idea
```
laurentmuller: The .idea must be ignored globally.

pluk77:
Why would you want to update the minimum PHP requirements and other dependencies? Branch 1.x is compatible with SF 4 and 5 which do not require PHP8, nor does it require some of the newer versions of the dependencies that you are proposing.

brouquet-company:
> Why would you want to update the minimum PHP requirements and other dependencies? Branch 1.x is compatible with SF 4 and 5 which do not require PHP8, nor does it require some of the newer versions of the dependencies that you are proposing.

I initially proposed updating the minimum PHP version and dependencies because I wanted to ensure this bundle would be compatible with Symfony 7.3.
However, taking a step back, I realize this might not be the right approach here. It introduces stricter requirements that break compatibility with older Symfony versions, while the 1.x branch was designed precisely to support Symfony 4 and 5 without enforcing PHP 8+.

So I agree it makes more sense to preserve the current compatibility in 1.x rather than raising the minimum requirements.

PR #244, opened 2025-11-04
Support for Symfony 8.x

michnovka:
Also updated tests while at it


michnovka:
Seems this requires much more work... I dont have the time now unfortunately :(

michnovka:
I am sorry, I give up. I cannot spend more time on this. I am not sure if the errors regarding native lazy objects mean that I just got the configuration wrong, or that it actually is incompatible with newer doctrine versions

ogizanagi:
Thanks for investigating, that's already precious. We'll get to that later

